PR #42, opened 2021-05-15
Add Arch Linux Installation info

LaurentFough:
PKG here: https://aur.archlinux.org/packages/iptvnator-bin/


4gray:
Thank you for your contribution!

PR #106, opened 2022-01-04
add support for loading the epg data based on tvg.id (if provided)

phartshorn:
This MR will attempt to match based on tvg id if provided in the playlist. If there's no tvg.id for a channel, the old behaviour will be used.

Notes: 
* I added a falsy check for the nameObj.value for my playlist has some names that are empty (I'm going to be looking at that next as the playlist works in other iptv apps)
* I also changed the callbacks for as they were, the callback functions weren't actually being called.

My System:
* macOS Monterey
* node v17.2.0


4gray (approved):

4gray:
Thank you for the contribution and for the callback fix!

PR #318, opened 2023-11-07
Update spanish translation

sguinetti:
Greetings and thank you for having accepted the PR earlier. This next is PR is for update es.json file with new strings.


4gray:
Muchas gracias :-)

PR #332, opened 2023-12-02
feat: support for setting channel HTTP referrer headers for HTML 5 Player

sapireli:
This commit introduces support for setting the HTTP referrer headers per channel. The referrer is the http-referrer as defined for each channel in the m3u8 playlist using the VLC format (#EXTVLCOPT:http-referrer=http://example.com/) This feature works currently only for the HTML5 player. Would love any insight on how to do it for the js one.

Updates were made to the html-video-player component for the HTML5 player, and to the logic in the CHANNEL_SET_USER_AGENT function in the main electron api.ts.

### DIFF
--- a/electron/api.ts
+++ b/electron/api.ts
@@ -169,14 +169,12 @@ export class Api {
             .on(
                 CHANNEL_SET_USER_AGENT,
                 (_event, args: { userAgent: string; referer?: string }) => {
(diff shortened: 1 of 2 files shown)


sapireli:
@4gray if you could please accept... many thanks

4gray:
Thank yor the contribution!

PR #349, opened 2024-01-14
feat: added support for setting the "origin" http header.

sapireli:
This commit adds an http origin header which is by default set to be the same as the referer without the trailing '/'. 
This is to support certain popular Eastern European IPTV providers.

A more elegant approach would be to define a origin header format for each channel specified in the m3u8 file. Perhaps in a future commit if i have the time.


4gray:
Thank you for the contribution!

PR #434, opened 2024-09-27
fix(vlc): path in quotes to allow path with spaces

RyPort:
Currently, paths with spaces (C:\Program Files\VideoLAN\VLC) don't work. This is the most common install directory of vlc on windows. I put the path in quotes when passing it to the child process to fix this


RyPort:
This seems to cause some other weirdness with the path input. Clsoing

windowshopr:
Any update on this?

RyPort:
> Any update on this?

I couldn't figure it out. Just install vlc to c:/vlc and you won't have any issues

Issue #435, opened 2024-09-27
Xtream Codes does not display EPG when using MPV or VLC as player


martymcflywa:
**Describe the bug**
Given Xtream Codes playlist and MPV or VLC player is used, when a live tv channel is played, then the EPG is not displayed.

**To Reproduce**
1. Log in to service using Xtream Codes
2. Go to Live Streams
3. Select a category
4. Select a channel in the sidebar

**Expected behavior**
The EPG for the channel should be displayed in the main container.

**Screenshots**
![Screenshot 2024-09-27 at 10 01 01 pm](https://github.com/user-attachments/assets/ce68a96b-366d-4a09-bc70-6bb6fd92207f)
In the screenshot above, the channel "10 HD" has been selected and is playing in an MPV window. However, the text in the main container continues to say "Please select channel". The channel EPG should be displayed in place of "Please select channel".

If it displayed the EPG, then it would behave similarly to when VideoJS is selected:
![Screenshot 2024-09-27 at 10 14 45 pm](https://github.com/user-attachments/assets/2ae9e7b6-0ab2-4133-91e8-f8748e1de50d)

**Desktop (please complete the following information):**
 - Electron
 - OS: MacOS
 - Browser Chrome
 - Version 129.0.6668.70

PR #436, opened 2024-09-27
fix(xtream): display epg when live tv channel is selected

martymcflywa:
Fixes https://github.com/4gray/iptvnator/issues/435

Set this.streamUrl so that child live-stream-layout displays the EPG.

Before:
Screenshot 2024-09-27 at 10 01 01 pm

After:
Screenshot 2024-09-27 at 10 02 26 pm

Apologies, this is the first time i've contributed to this repository so i'm not 100% sure which branch this should be targeting. If this change is ok, please let me know if i need to select a different branch target.

Appreciate your work on this great app by the way. v0.16.0 is looking very nice!


4gray:
Just redirected the PR to `electron` branch, thanks for the feedback and the fix 👍

PR #474, opened 2024-11-28
Fix for Dependency Injection Issue in DataFactory

abdullahceylan:
This PR resolves a critical issue where NgxIndexedDBService was being injected into PwaService instead of HttpClient, causing runtime errors when making HTTP requests. The issue was traced to an incorrect dependency injection order in the DataFactory function.

Screenshot 2024-11-28 at 15 14 06


4gray:
Thanks for the investigation and fix!

PR #538, opened 2025-03-16
Add stream format selection

p20f:
Introduce a configuration option to select between TS and M3U8 stream formats, along with Italian translations for these options.


4gray:
Hi, thank you for the contribution!
Just two things please:
1. It would be good to mention, that the stream format selection is related only to xtream.
2. Switch the default stream format value to m3u8, like it was before.

4gray:
@p20f thank you for the contribution!

PR #547, opened 2025-03-31
Longwaiting AC3 support

Kif-Fando:
Latest version of HLS already support AC3 audio and maybe something else.
Tested on MacOS, all works well


4gray:
Thank you!

PR #553, opened 2025-04-09
Group sorting logic

Kif-Fando:
Change the logic sort groups:
before:
Screenshot 2025-04-09 at 19 03 54
after:
Screenshot 2025-04-09 at 19 04 50

Before:
1. group1
10. group 10
2. group 2

After:
1. group 1
2. group 2
10. group 10

P.S. BTW upgrading the version of HLS library


4gray:
Looks good, thank you!

PR #902, opened 2026-05-07
feat(xtream): fall back to uploaded XMLTV when provider has no EPG

weeco:
Live TV in Xtream playlists shows "No program information available" whenever the provider's `get_short_epg` returns nothing, even when there are working XMLTV URLs in Settings -> EPG. The XMLTV pipeline already populates `epg_programs` but only the M3U module was reading from it.

This wires those uploads up as a fallback for Xtream too. When the provider returns nothing for a channel with an `epg_channel_id`, the channel is looked up in the local table and used instead. A settings toggle flips the order for users whose curated XMLTV is better than the provider's auto guide.

Bildschirmfoto 2026-05-07 um 08 24 56

### DIFF
--- a/apps/web/src/app/settings/settings-epg-section.component.html
+++ b/apps/web/src/app/settings/settings-epg-section.component.html
@@ -89,6 +89,19 @@
         </div>
     </div>
 
+    <div class="setting-item">
+        <div class="setting-item__meta">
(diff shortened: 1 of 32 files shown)


weeco:
@4gray I addressed the PR feedback, but can't trigger a re-view here I believe.

4gray:
> @4gray I addressed the PR feedback, but can't trigger a re-view here I believe.

@weeco All good, thank you for contributing! I’ll take a look over the weekend 👍

PR #903, opened 2026-05-07
Fix Xtream section search across categories

sapireli:
## What changed

This fixes workspace header search inside Xtream section root views without changing category-local search behavior.

The intended behavior is:

- On the Movies root view, search across all movie categories.
- On the Series root view, search across all series categories.
- On the Live TV root view, search across all live categories.
- Once a specific category is selected, search only inside that selected category.

Before this change, search from Movies and Series root views could miss results because the section/category filtering path was not explicit enough. Live TV had one extra problem: even when the search state contained the right query, the Live root layout still rendered the "select category" empty state after loading instead of showing the searchable channel list.

## Root cause

Xtream section lists are driven by selected content type, selected category, and the workspace search query. The base list has to be chosen before applying the query:

- If a category is selected, the base list should be that category's items.
- If no category is selected, the base list should be all items for the selected section type.

Live TV also uses a dedicated player layout rather than the Movies/Series catalog grid. That layout only showed the channel list when a category was selected or content was still loading, so Live root search had no visible result list once loading finished.

## Fix

The Xtream selection store now builds the search list from the right boundary first:

- selected category -> filter that category by the search term
- no selected category -> filter the full selected section by the search term

Live TV routes now sync the workspace query into the Xtream section search state, and the Live root layout renders the channel list whenever a search term is active.

## Validation

- pnpm nx test portal-xtream-data-access --runTestsByPath libs/portal/xtream/data-access/src/lib/stores/features/with-selection.feature.spec.ts
- pnpm exec jest -c libs/workspace/shell/feature/jest.config.ts --runTestsByPath libs/workspace/shell/feature/src/lib/workspace-shell/services/workspace-shell.facade.spec.ts
- pnpm exec jest -c libs/portal/xtream/feature/jest.config.ts --runTestsByPath libs/portal/xtream/feature/src/lib/live-stream-layout/live-stream-layout.component.spec.ts
- Prettier check on touched files

Also built unsigned local macOS DMGs for manual testing:

- dist/executables/iptvnator-0.22.0-mac-x64.dmg
- dist/executables/iptvnator-0.22.0-mac-arm64.dmg

Note: portal-xtream-data-access lint still reports existing circular dependency errors outside this change.


4gray:
@sapireli Thanks for contributing! I found a small issue with click handling in the internal player after using search, so I fixed it on top of this PR and added regression coverage.

WolfganP:
Thx guys for this patch. The live TV search logic was driving me crazy, but this enhancement really works wonders on search usability.
Thx a lot!

PR #920, opened 2026-05-10
fix(playback): avoid redundant VideoJS source resets

### LINKED ISSUE #608: VideoJS player hangs / freezes shortly after stream starts
Some HLS streams will freeze within 1 minute when the selected player is VideoJS ([example](https://freespeech.akamaized.net/hls/live/2016712/live1/playlist.m3u8))

```
CODECS = avc1.640029, mp4a.40.2

BANDWIDTH = 2604980, 1280 x 720
BANDWIDTH = 378651,   320 x 180
BANDWIDTH = 694383,   480 x 270
BANDWIDTH = 973268,   640 x 360
BANDWIDTH = 1920142,  960 x 540

BANDWIDTH = 217963, AUDIO = aud-192000
```

4gray:
## Summary
- Prevent VideoJS from resetting playback when Angular provides a new `options` object for the same stream source.
- Add focused regression coverage for equivalent source updates and real source changes.
- Add a `ui-playback` Nx test target so playback component specs can run directly.

## Root cause
`VjsPlayerComponent.ngOnChanges` treated every `options` reference change as a source change and called `player.src(...)`. Inline Angular bindings and other callers can provide a fresh options object with the same HLS source, which restarts VideoJS and can disrupt long-running HLS playback.

## Testing
- [x] `pnpm nx test ui-playback --runInBand`
- [x] `pnpm exec eslint libs/ui/playback/src/lib/vjs-player/vjs-player.component.ts libs/ui/playback/src/lib/vjs-player/vjs-player.component.spec.ts`
- [x] `pnpm nx build web`
- [ ] `pnpm nx lint ui-playback` currently fails on existing module-boundary circular dependency errors outside this change.

Closes #608

### DIFF
--- a/libs/ui/playback/project.json
+++ b/libs/ui/playback/project.json
@@ -6,6 +6,21 @@
     "projectType": "library",
     "tags": ["scope:shared", "domain:playback", "type:ui"],
     "targets": {
+        "test": {
+            "executor": "nx:run-commands",
+            "outputs": ["{workspaceRoot}/coverage/{projectRoot}"],
+            "options": {
+                "command": [
+                    "node",
+                    "./tools/testing/run-web-esm-lib-tests.mjs",
+                    "libs/ui/playback/src"
+                ],
+                "env": {
+                    "NODE_OPTIONS": "--experimental-vm-modules"
+                },
+                "forwardAllArgs": true
+            }
+        },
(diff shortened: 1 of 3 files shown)


4gray:
@greptileai

PR #924, opened 2026-05-10
fix(electron): resolve macOS external player app paths

4gray:
## Summary
- Resolve custom macOS MPV/VLC `.app` bundle paths to the executable inside `Contents/MacOS` before spawning the external player.
- Keep already-correct executable paths unchanged, including `/Applications/mpv.app/Contents/MacOS/mpv`.
- Add regression coverage for MPV, VLC, and backward-compatible executable paths.

Fixes #918

## Changes
- Add macOS-only `.app` bundle path normalization in the external player launch context.
- Preserve Linux, Windows, and Flatpak launch behavior.
- Extend `player.events.spec.ts` with bundle-path and existing executable-path cases.

## Testing
- [x] `pnpm nx test electron-backend --testFile=apps/electron-backend/src/app/events/player.events.spec.ts --runInBand`
- [x] `pnpm run typecheck:backend`
- [ ] `pnpm nx lint electron-backend` currently fails on an existing `@nx/enforce-module-boundaries` error at `apps/electron-backend/src/main.ts:2`, unrelated to this change.

PR #929, opened 2026-05-12
fix(settings): clarify external player path hints

4gray:
## Summary
- Clarify MPV/VLC path settings after macOS app-bundle resolution from #924.
- Add an IINA-specific macOS tip that points users to executable paths instead of `/Applications/IINA.app`.
- Update docs and release-note copy so IINA launch is described as best-effort, while IPTVnator controls and position polling remain MPV/VLC-specific.

## Changes
- Adds MPV-compatible-player guidance in Settings playback UI.
- Updates MPV/VLC path descriptions across locale files.
- Adds regression coverage for MPV/VLC hint visibility.

## Testing
- [x] `pnpm nx test web`
- [x] `git diff --check`

## Docs
- Updated `README.md`, `docs/architecture/embedded-inline-playback.md`, and the v0.21 release note.
- Wiki export skipped locally because `IPTVNATOR_WIKI_VAULT` is not set.


4gray:
@greptileai

PR #934, opened 2026-05-13
fix(playback): handle IPTV proxy URLs without file extensions

tonioriol:
## Problem

IPTV proxy servers (Acexy, Xtream Codes, etc.) use query-string URLs like:
```
https://proxy.example.com/ace/getstream?infohash=abc&token=xyz
```

These return raw MPEG-TS streams but have no file extension in the URL path. `getExtensionFromUrl()` returned garbage for these URLs (e.g. `com/ace/getstream` from splitting on dots across path segments), causing all four player backends to misroute the stream — typically to HLS.js, which fails because the response is raw MPEG-TS, not an HLS manifest.

**Result:** "The media could not be loaded" / "No compatible source" for all built-in players when using IPTV proxy URLs.

## Root Cause

`getExtensionFromUrl()` splits the entire URL path on dots instead of extracting the extension from the last path segment only.

## Fix

**1. Fix `getExtensionFromUrl()`** (the root cause):
- Extract extension from the last path segment only (after the final `/`)
- Return `undefined` when there's no dot-separated extension

**2. Handle `undefined` extension in all 4 player backends:**
- HTML5 player, ArtPlayer, VideoJs, web-player-view: treat `undefined` extension as MPEG-TS (the most common format for live IPTV proxy streams)

## Changes

| File | Change |
|------|--------|
| `libs/shared/m3u-utils/src/lib/playlist.utils.ts` | Fix `getExtensionFromUrl()` to parse last path segment |
| `libs/ui/playback/.../html-video-player.component.ts` | Route `!extension` to mpegts.js |
| `libs/ui/playback/.../art-player.component.ts` | `isLive` + `getVideoType()` handle `!extension` |
| `libs/ui/playback/.../vjs-player.component.ts` | `isMpegTsSource()` handles `!extension` |
| `libs/ui/playback/.../web-player-view.component.ts` | MIME type `video/mp2t` for `!extension` |

**5 files changed, 24 insertions, 7 deletions.**

## Testing

Tested with Acexy IPTV proxy URLs (`/ace/getstream?infohash=...`) — all four built-in players now correctly use mpegts.js for these extensionless URLs. No behavior change for URLs with recognized extensions.

### DIFF
--- a/libs/shared/m3u-utils/src/lib/playlist.utils.spec.ts
+++ b/libs/shared/m3u-utils/src/lib/playlist.utils.spec.ts
@@ -1,5 +1,5 @@
 import { Channel, Playlist } from 'shared-interfaces';
-import { aggregateFavoriteChannels } from './playlist.utils';
+import { aggregateFavoriteChannels, getExtensionFromUrl } from './playlist.utils';
 
 function createChannel(id: string, url: string, name = id): Channel {
     return {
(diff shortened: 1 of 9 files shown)


4gray:
Thanks for the contribution. I added regression coverage for URL extension parsing and kept the extensionless IPTV proxy fallback, but preserved query-declared stream formats (for example `?extension=m3u8` / `?extension=ts`) before falling back to MPEG-TS. This should avoid regressing signed/proxy HLS URLs while keeping the Acexy/Xtream TS case fixed.

4gray:
@greptileai

tonioriol:
@4gray great, i was a bit unsure about the hard fallback, that makes more sense!

PR #941, opened 2026-05-15
fix(stalker): preserve explicit portal identity

### LINKED ISSUE #860: Not to use default deviceID if not provided
I have noticed that when we authenicate for stalker portal, it used the default deviceid1 generated inside stalker-session.service.ts

i would suggest not to use any deviceID if not provided by user. As i use the same account in STBEMU app, and we don't use any deviceID in there, but once i used the account with IPTVNATOR, it will send this deviceID to service provided which will make it stop working on STBEMU.

Note: I have noticed couple of things inside stalker-session.service.ts
1.  Inside ensureToken, we are calling `authenticate`, where we are not passing in any values for deviceID, So this will be auto generate everytime, even though user have provided it in configuration??
```
const authPromise = (async () => {
            try {
                const { token } = await this.authenticate(
                    playlist.portalUrl,
                    playlist.macAddress,
                    serialNumber
                );
                this.setCachedToken(playlist._id, token);
                return { token, serialNumber };
            } finally {
                // Clean up pending promise regardless of success/failure
                this.pendingAuth.delete(playlist._id);
            }
        })();
```

2. Inside `authenticate` we call `getProfile`, my recommendation is not to autogenerate the deviceID1 OR have a new configuration, like checkbox asking user to generate deviceID based on MACAddress. DeviceID1 not provided and generate from mac not selected, don't send deviceID1/deviceID2 in request/params

Please let me know if you need some help with this

### LINKED ISSUE #927: Stalker Parameters entered are not forwarded to API
**Describe the bug**
Stalker optional parameters (SN, Device IDs & Signatures), when entered, are not being forwarded to the API, and the SN parameter, whether empty or filled, is being generated by the Application. Due to which application is giving error "unable to load categories"? ( Actually due to "Device conflict - device_id mismatch")
This is not happening in all the Stalker portals cases, only with portals with strict SN & Device ID requirements.

**To Reproduce**

1. Go to Add playlist ==> Stalker ==> Enter Parameters like URL, MAC, SN & Device ID
2. Click on 'Add'
3. Open a HTTP debugger and scan the URL Parameters being forwarded to API
4. You will find - 1. SN changed, 2. Device IDs changed to Zeros.

**Expected behavior**
If SN and Device ID values are entered, then same to be forwarded to the API. Or an option to either generate it through the app or leave it blank. ( As some portals don't require these parameters to work)

**Screenshots**

**Laptop**
-   IPTVnator application
-   OS: Windows 11
-   Version 0.21
-   Feature: Stalker Portal

**Additional context**
Option to add Stalker without SN and Device IDs ( if not entered gives error and doesn't save parameters)

4gray:
## Summary
- Fix Stalker/Ministra full-portal auth so only explicit user-provided identity fields are sent.
- Preserve provided serial/device/signature values across import, auth, token refresh, retries, API requests, and same-origin playback headers.
- Treat blank identity fields and the legacy default serial `BEDACD4569BAF` as absent at runtime.

## Changes
- Added shared Stalker identity normalization and wired it through renderer data-access and Electron request handling.
- Updated import persistence to save only trimmed canonical `stalker*` identity fields and pass signatures into initial auth.
- Extracted Electron Stalker identity/header/cookie handling into a pure helper with focused tests.
- Added playback header normalization so legacy serials are not sent as `SN` or serial-derived `__cfduid`.
- Documented the MAC-only default and explicit identity passthrough policy.

## Testing
- [x] `NX_DAEMON=false pnpm nx test portal-stalker-data-access --runInBand --skipNxCache`
- [x] `NX_DAEMON=false pnpm nx test playlist-import-feature --runInBand --skipNxCache`
- [x] `NX_DAEMON=false pnpm nx test electron-backend --runInBand --skipNxCache`
- [x] `NX_DAEMON=false pnpm nx test portal-stalker-feature --runInBand --skipNxCache`
- [x] `pnpm run typecheck:ci`
- [x] `git diff --check`

Fixes #860
Fixes #927

### DIFF
--- a/apps/electron-backend/src/app/events/stalker-identity.spec.ts
+++ b/apps/electron-backend/src/app/events/stalker-identity.spec.ts
@@ -1,0 +1,95 @@
+import {
+    LEGACY_DEFAULT_STALKER_SERIAL,
+    buildStalkerIdentityRequestContext,
+} from './stalker-identity';
+
+describe('buildStalkerIdentityRequestContext', () => {
+    const macAddress = '00:1A:79:AA:BB:CC';
+
+    it('does not add SN header or force get_profile SN fields when serial is absent', () => {
+        const context = buildStalkerIdentityRequestContext({
+            macAddress,
+            params: {
+                type: 'stb',
+                action: 'get_profile',
+                sn: 'STALE-SN',
+                metrics: JSON.stringify({
+                    mac: macAddress,
+                    sn: 'STALE-SN',
+                }),
+            },
+        });
+
+        expect(context.effectiveSerialNumber).toBeUndefined();
+        expect(context.headers).not.toHaveProperty('SN');
+        expect(context.cookieString).not.toContain('__cfduid=');
+        expect(context.requestParams).not.toHaveProperty('sn');
+        expect(
+            JSON.parse(String(context.requestParams.metrics))
+        ).not.toHaveProperty('sn');
+    });
+
+    it('preserves a provided serial exactly in headers, params, and metrics', () => {
+        const context = buildStalkerIdentityRequestContext({
+            macAddress,
+            serialNumber: 'CustomSn123',
+            params: {
+                type: 'stb',
+                action: 'get_profile',
+                metrics: JSON.stringify({
+                    mac: macAddress,
+                }),
+            },
+        });
+
+        expect(context.effectiveSerialNumber).toBe('CustomSn123');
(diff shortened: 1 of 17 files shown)


4gray:
@codex review

4gray:
@greptileai

rupisaini123:
@4gray, when is this going to be part of release?

PR #963, opened 2026-05-21
ci: cache embedded mpv runtime

4gray:
## Summary
- Add exact-keyed GitHub Actions cache restore/save around the staged macOS embedded MPV runtime used for temporary PR/master artifact testing.
- Keep tagged release builds on the pinned source-build path while treating cache hits as a CI speed optimization only.
- Document the runtime cache policy and the release-grade artifact/attestation boundary.

## Validation
- [x] Parsed `.github/workflows/build-and-make.yaml` with Ruby YAML loader.
- [x] Ran `git diff --check` for the workflow and embedded MPV docs.
- [x] Verified the local cache-key shell snippet produces a key.
- [x] Ran `gh workflow view build-and-make.yaml`.

## Notes
- The `.plans/2026-05-22-embedded-mpv-ci-runtime-cache.md` handoff plan was saved locally as requested, but `.plans/` is ignored by this repository and is not included in the PR.


4gray:
@greptileai

PR #1033, opened 2026-06-09
fix(xtream): open correct channel and category from Ctrl+F live search

ilyesbrh:
## Summary

- **Component reuse bug**: when already on the `/live` route, Angular reused `LiveStreamLayoutComponent` without re-running the constructor, so `openXtreamLiveItemId` was never read from history state on re-navigation. Fixed by subscribing to `NavigationEnd` and re-checking the state on every navigation.
- **Wrong content type guard**: the auto-open effect could fire before `syncRouteState` set `selectedContentType` to `'live'`, causing the channel to be looked up in VOD streams, not found, and the pending ID immediately cleared. Fixed by guarding on `selectedContentType() === 'live'` before processing.
- **Category filter blocking lookup**: `getVisibleChannels()` only returned channels in the currently selected category, so a channel from a different category was never found. Switched to `getAllLiveStreams()` (raw `liveStreams` signal) for the auto-open lookup.
- **Sidebar not updating**: after opening the channel, `selectedCategoryId` is now set to the channel's category so the sidebar shows the correct category and highlights the playing channel.

## Test plan

- [ ] Open an Xtream playlist and navigate to the Live TV tab with a category selected
- [ ] Press Ctrl+F, search for a live channel in a **different** category, click it → channel plays and sidebar updates to its category
- [ ] Press Ctrl+F while already on the Live TV tab, click a live channel → channel plays (component reuse case)
- [ ] Press Ctrl+F while on the VOD tab, click a live channel → navigates to Live TV and plays the channel

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/libs/portal/xtream/feature/src/lib/live-stream-layout/live-stream-layout.component.ts
+++ b/libs/portal/xtream/feature/src/lib/live-stream-layout/live-stream-layout.component.ts
@@ -2,6 +2,7 @@ import { NgTemplateOutlet } from '@angular/common';
 import {
     ChangeDetectionStrategy,
     Component,
+    DestroyRef,
     HostListener,
     computed,
     effect,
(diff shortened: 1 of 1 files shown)


ilyesbrh:
Closing to reopen from correct account.

PR #1040, opened 2026-06-10
fix(epg): dedupe concurrent fetches and await worker termination

4gray:
## Problems

1. **Duplicate workers for the same URL.** Two concurrent `fetchEpgFromUrl` calls for the same URL (e.g. a retry racing a slow first fetch) spawned two workers that both parsed and wrote the same EPG data. The second worker overwrote the first one's entry in the `workers` map, so the first worker leaked from the map's perspective.
2. **Fire-and-forget `worker.terminate()`.** Every settle path (timeout, complete, error, clear) called `terminate()` without awaiting it. A terminated-but-still-exiting worker can keep holding the SQLite lock and block the next EPG operation that starts immediately after the promise settles.

## Fix

- `fetchEpgFromUrl` keeps an `inFlightFetches` map and returns the existing in-flight promise for a URL instead of spawning a competing worker; the entry is cleaned up when the fetch settles.
- New `terminateWorker()` helper awaits `worker.terminate()` (logging failures). All settle paths now resolve/reject only after the worker thread has actually exited. The `settled` guard is set before termination starts, so the worker's own `exit` event cannot hijack the outcome (e.g. a timeout still rejects with the timeout error, not "worker exited unexpectedly").

## Test impact

Extended `epg.events.spec.ts` with three regression tests (all fail on the old behavior):
- concurrent fetch of the same URL reuses the single in-flight worker;
- a failed fetch cleans up so a retry spawns a fresh worker;
- a timed-out fetch terminates the worker and still rejects with the timeout error even when the worker emits `exit` during termination.

`pnpm nx test electron-backend` — 23 suites, 254 tests passed. `pnpm nx lint electron-backend` — clean. Docs: no updates needed (internal lifecycle fix).

### DIFF
--- a/apps/electron-backend/src/app/events/epg-worker.service.ts
+++ b/apps/electron-backend/src/app/events/epg-worker.service.ts
@@ -21,6 +21,7 @@ interface EpgWorkerMessage {
 export class EpgWorkerService {
     private readonly fetchedUrls = new Set<string>();
     private readonly workers = new Map<string, Worker>();
+    private readonly inFlightFetches = new Map<string, Promise<void>>();
 
     constructor(
         private readonly loggerLabel = '[EPG Events]',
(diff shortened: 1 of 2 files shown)


4gray:
@greptileai

PR #1030, opened 2026-06-07
feat(playback): add embedded mpv series navigation

4gray:
## Summary
- Add current `SxxExx` episode metadata to the inline player header for Xtream and Stalker series playback.
- Add embedded MPV previous/next episode controls plus default-on current-season autoplay.
- Extend embedded MPV lifecycle with `ended` so autoplay only reacts to successful EOF, not close/dispose/error.

## Changes
- Propagates embedded MPV `ended` status from native addon through Electron services and playback components.
- Adds shared series navigation metadata/state for portal inline playback.
- Wires Xtream and Stalker series detail views to compute previous/next only within the current season.
- Adds focused regression coverage for MPV EOF, one-shot ended emission, nav boundaries, and Stalker `is_series=1` lazy-season behavior.
- Updates embedded MPV and Stalker architecture docs.

## Testing
- [x] `pnpm nx test ui-playback`
- [x] `pnpm nx test portal-xtream-feature`
- [x] `pnpm nx test portal-stalker-feature`
- [x] `pnpm nx test electron-backend`
- [x] `pnpm nx run web-e2e:e2e-ci--src/stalker.e2e.ts`
- [x] `git diff --check`
- [x] `pnpm run serve:backend:embedded-mpv` built the Homebrew dev native addon and launched Electron without TypeScript errors
- [x] `agent-browser` renderer boot check on `http://localhost:4200` with no page errors

## Notes
- `pnpm nx run web-e2e:e2e-ci--src/xtream.e2e.ts` still has one existing fixture/API expectation failure: `src/xtream.e2e.ts:218` expects 3 seasons and receives 2. The changed inline playback flow is covered by unit tests.
- Electron CDP on port `9222` exposed `about:blank` instead of the application renderer target in this local run, so real embedded MPV series playback was not manually exercised through the app window.
- Wiki export was skipped because `IPTVNATOR_WIKI_VAULT` is not set locally.

PR #1041, opened 2026-06-10
fix(playback): harden embedded MPV polling loop and IPC error visibility

4gray:
## Problems

Two robustness gaps in the embedded MPV code from #1030/#1031:

1. **Unguarded polling loop.** The 500 ms session polling interval in `EmbeddedMpvNativeService` called `refreshSession()` without a try-catch. A throwing addon call (`getAddon()` after the addon becomes unavailable, or a native `getSessionSnapshot` failure) escaped the interval callback as an `uncaughtException` in the main process — repeated on every tick while the timer keeps running.
2. **Invisible IPC errors.** All `EMBEDDED_MPV_*` handlers forwarded service calls without any error handling, unlike every other events module (cf. `playlist.events.ts` log-and-rethrow). Since the renderer deliberately swallows these rejections (`guardIpc` resyncs from the next snapshot), addon-side errors left no trace anywhere.

## Fix

- Polling: wrap each `refreshSession()` in try-catch; log only the first consecutive failure (no log flooding at poll rate) and resume normal session updates once the addon recovers.
- IPC: route all 16 registrations through a `handleEmbeddedMpv()` wrapper that logs the failing channel in the main process and rethrows, preserving the renderer-visible rejection.

## Test impact

- `embedded-mpv-native.service.spec.ts`: new test — three poll ticks with a throwing snapshot do not escape the interval, failure is logged once, updates resume after recovery (fails on old behavior).
- New `embedded-mpv.events.spec.ts`: handlers forward arguments/results, and a throwing service call is logged and rethrown.
- `pnpm nx test electron-backend` — 24 suites, 254 tests passed. `pnpm nx lint electron-backend` — clean.
- Docs: no updates needed (internal robustness fix, no behavior contract change).

### DIFF
--- a/apps/electron-backend/src/app/events/embedded-mpv.events.spec.ts
+++ b/apps/electron-backend/src/app/events/embedded-mpv.events.spec.ts
@@ -1,0 +1,82 @@
+jest.mock('electron', () => ({
+    ipcMain: {
+        handle: jest.fn(),
+    },
+}));
+
+const mockEmbeddedMpvService = {
+    getSupport: jest.fn(),
+    setPaused: jest.fn(),
+};
+
+jest.mock('../services/embedded-mpv-native.service', () => ({
+    EmbeddedMpvNativeService: class {},
+    embeddedMpvNativeService: mockEmbeddedMpvService,
+}));
+
+import { ipcMain } from 'electron';
+import {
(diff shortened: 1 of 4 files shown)


4gray:
@greptileai

PR #1042, opened 2026-06-10
feat(ui): custom title bar with window controls for Windows and Linux

4gray:
## Summary

Extends the existing macOS custom title bar (`titleBarStyle: 'hidden'` + native traffic lights) to **Windows and Linux** with renderer-drawn, theme-aware window controls, and upgrades **Electron to 41** so frameless windows get proper decorations (drop shadow, resize boundaries) on Linux/Wayland.

### Window controls (min / max-restore / close)

- `titleBarStyle: 'hidden'` now applies on win32/linux too; `frame` stays untouched so native resize borders and OS snapping keep working.
- New `WINDOW:*` IPC contract (`MINIMIZE`, `TOGGLE_MAXIMIZE`, `CLOSE`, `GET_STATE` + `STATE_CHANGED` push) in `window.events.ts`, resolved from the sender WebContents; close goes through `win.close()` so window-bounds persistence still runs. State pushes keep the maximize/restore glyph correct for OS-triggered changes (snap, double-click, F11); controls hide in fullscreen.
- `WindowControlsComponent` (`libs/ui/components`) mounts once in `app-root` as a **manual popover** so it lives in the browser top layer — CDK overlays render as top-layer popovers where no z-index competes — keeping the buttons clickable above dialog backdrops and the multi-EPG overlay, exactly like macOS traffic lights. It re-asserts itself whenever another popover opens.
- Theme-aware purely via CSS vars (`--app-on-surface`, `--app-hover-overlay`) with the classic Windows-red close hover; drag regions get right padding via a body-level `frameless-platform` class. Gated by `RuntimeCapabilitiesService.usesCustomWindowControls` — PWA and macOS never mount it.

### Electron 41 upgrade (Linux window decorations)

Hiding the title bar removed the WM-drawn shadow/rounded corners on Linux. Electron only draws client-side decorations on native Wayland, and frameless-window CSD (GTK drop shadow + extended resize boundaries) landed in **Electron 41** — before that frameless windows are plain rectangles. Verified side-by-side on Ubuntu/Wayland: 39 is flat, 41 is decorated.

- `electron` `^39.8.5` → `^41.7.2` (Wayland auto-detected since 38.2; X11 sessions remain undecorated like other frameless Electron apps; Windows keeps DWM shadow + Win11 rounded corners). Reviewed 40/41 breaking changes — only the renderer `clipboard` deprecation, unused here.
- `better-sqlite3` pinned to exactly `12.9.0`: last release with prebuilds for **both** Node 20 (ABI 115, Jest) and Electron 41 (ABI 145, runtime); 12.10.0 dropped the Node 20 prebuilds.
- pnpm override `node-abi` → 3.92.0 so `electron-builder install-app-deps` can map Electron 41 to ABI 145.

### Known limitations

- No Windows 11 Snap Layouts flyout on the custom maximize button (inherent to non-native caption buttons).
- Rounded corners for frameless Linux windows are not yet supported upstream (on Electron's roadmap).

Docs: new "Window Chrome And Custom Title Bar" section in `docs/architecture/workspace-shell.md` covering the IPC contract, top-layer placement rationale, Linux decoration matrix, and the dependency pins.

## Test plan

- [x] `pnpm nx test electron-backend` — 270/270 (incl. new `window.events.spec.ts`)
- [x] `pnpm nx test components` — 67/67 (incl. new `window-controls.component.spec.ts`)
- [x] `pnpm nx test web` / `pnpm nx test services` — 64/64, 85/85
- [x] `pnpm nx run electron-backend-e2e:e2e-ci--src/window-controls.e2e.ts` — 4/4 under Electron 41 (real maximize/minimize round-trips)
- [x] `smoke.e2e.ts` 3/3, `search.e2e.ts` 16/16 under Electron 41 (frameless launch + DB via rebuilt better-sqlite3)
- [x] Lint clean across touched projects
- [x] Manual CDP verification with agent-browser on Ubuntu/Wayland: light + dark theme button colors, red close hover, maximize/restore glyph sync, controls clickable above modal backdrop, minimize, close persists window bounds, Electron 39 vs 41 shadow comparison


4gray:
Pushed 315d11cc addressing the CI failure and both review points:

- **Linux CI E2E failure**: the three failing tests assert real window-manager state (maximize/minimize), but the ubuntu runner drives Electron under xvfb with no WM, so that state never materializes. Those three assertions are now skipped on Linux CI only — Windows CI, macOS, and local Linux desktop runs keep the coverage. The button-render test still runs everywhere.
- **`WINDOW:TOGGLE_MAXIMIZE` stale return**: the handler now returns the *requested* state instead of re-reading `isMaximized()` immediately after the async maximize/unmaximize call; the `WINDOW:STATE_CHANGED` push remains the authoritative update.
- **Dead IPC on macOS**: `attachWindowStateEvents` now early-returns on darwin, since the custom controls (the only subscriber) never mount there.

4gray:
@greptileai review

4gray:
Re the two follow-up observations:

- **Capability gate** — applied in 21bd0980: `usesCustomWindowControls` now requires `getWindowState` and `onWindowStateChange` too, so a partial bridge can't mount controls whose glyph would never sync.
- **`bootstrapWindowEvents()` as empty stub** — declining: module-level `ipcMain.handle` registration with an empty bootstrap method is the established pattern in this codebase (`electron.events.ts`, `settings.events.ts`, `xtream.events.ts`, `stalker.events.ts` all do exactly this); only some event classes register inside the method. `window.events.ts` was deliberately modeled on `electron.events.ts`.

4gray:
@greptileai

PR #1050, opened 2026-06-12
fix(playback): detect embedded mpv keep-open eof

4gray:
## Summary
- Fix embedded MPV series autoplay when mpv reaches EOF while `keep-open=yes` keeps the file loaded.
- Treat mpv's `eof-reached` property as the same successful `ended` session status used by the renderer autoplay contract.
- Keep MPV playlist/manifest redirects in `loading` instead of briefly falling back to `idle`.
- Extend native-source regression coverage and document the EOF/redirect session contract.

## Root Cause
Embedded MPV sessions run with `keep-open=yes`. On macOS this can leave mpv paused at EOF while the file remains loaded, so relying only on `MPV_EVENT_END_FILE` left the renderer in a paused-at-end state. Because Angular only autoplays the next episode after `status: ended`, series playback stopped instead of advancing.

## Changes
- Observe `eof-reached` in the macOS native addon and map true EOF to `SessionStatus::Ended`.
- Align the Windows/Linux shared native backend with the same `ended` mapping.
- Handle `MPV_END_FILE_REASON_REDIRECT` as `loading` on macOS and Windows/Linux so redirect-heavy streams do not expose a transient `idle` state.
- Add regression checks for EOF, redirect, and `keep-open`/`eof-reached` handling.
- Update `docs/architecture/embedded-mpv-native.md` with the session-end contract.

## Testing
- [x] `pnpm nx test electron-backend --runInBand --skip-nx-cache --testPathPatterns=embedded-mpv-native-source.spec.ts` (1 suite / 11 tests)
- [x] `pnpm nx test electron-backend --runInBand --skip-nx-cache` (26 suites / 272 tests)
- [x] `pnpm embedded-mpv:build-native:homebrew`
- [x] `git diff --check`
- [x] Manual Electron + agent-browser verification: Stalker `S01E07` advanced to `S01E08` after EOF with the next button disabled on the final episode.

Docs updated: `docs/architecture/embedded-mpv-native.md`. Wiki export skipped because `IPTVNATOR_WIKI_VAULT` was not configured.

### DIFF
--- a/apps/electron-backend/native/src/embedded_mpv.mm
+++ b/apps/electron-backend/native/src/embedded_mpv.mm
@@ -1136,6 +1136,13 @@ void runEventLoop(const std::shared_ptr<Session>& session)
                     endFile->reason == MPV_END_FILE_REASON_EOF &&
                     session->running.load()) {
                     session->snapshot.status = SessionStatus::Ended;
+                } else if (
+                    endFile &&
+                    endFile->reason == MPV_END_FILE_REASON_REDIRECT &&
+                    session->running.load()) {
+                    session->snapshot.status = SessionStatus::Loading;
+                    session->snapshot.error.clear();
+                    session->loadedPath = false;
                 } else if (session->running.load()) {
                     session->snapshot.status = SessionStatus::Idle;
                 }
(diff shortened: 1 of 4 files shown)


4gray:
@greptileai

PR #1091, opened 2026-06-25
feat(playback): add Ferrite (WASM/WebCodecs) inline player for software HEVC

multiduplikator:
Adds **Ferrite** as a fourth inline web player — a canvas + WebAssembly player ([`ferrite.js`](https://www.npmjs.com/package/ferrite.js), an mpegts.js-style drop-in) that software-decodes **HEVC** where the browser/PWA has no hardware HEVC decoder (and uses WebCodecs when it does). Handles **live and VOD**, rendering to a `<canvas>` with WebAudio.

### Why
On a machine/browser with no hardware HEVC decoder, `<video>`/MSE and the existing inline players can't play HEVC. Ferrite carries its own FFmpeg subset compiled to WASM and decodes it on worker threads, fast enough for live 4K — covering the gap mpegts.js can't (it's bound to MSE/`<video>`).

### What Ferrite brings (beyond decoding HEVC)
Capabilities that matter for broadcast/IPTV content:
- **Broadcast-grade video** — 10/12-bit decode; HDR (PQ/HLG, BT.2020) tone-mapped to SDR; correct per-stream color (BT.601/709/2020 + limited/full range); anamorphic SAR→DAR (1440×1080 4:3-pixel → 16:9); and software **bwdif deinterlacing** for interlaced broadcast.
- **Smooth playback** — an audio-master clock with a PLL-paced present loop, gentle live latency catch-up (≤1.05× toward the edge), and a display-cadence engine that stays smooth at any refresh (50-on-60, 25-on-75…). Under sustained load it degrades gracefully (skip non-reference frames → skip in-loop deblock → halve present rate) to keep audio and motion intact, recovering automatically.
- **Resilience for flaky streams** — reconnect-with-backoff at clean boundaries, an adaptive silence watchdog that re-opens a stalled-but-open socket, WebCodecs decoder self-heal (rebuild → software fallback), and state-machine teardown with no resource creep across play/stop.
- **Robust mid-stream** — resolution changes, cross-codec failover, and sparse parameter sets handled without a reload.
- **Architecture** — split-realm pipeline: a decode worker, a present worker (owns the OffscreenCanvas + WebGL2 + present loop), and a thin main-thread facade (audio + events) — so decode and present never contend.

### Integration
- New `VideoPlayer.Ferrite`, selectable in **Settings → Player** and the workspace command palette. `web-player-view` routes live & VOD to it when selected, and falls back to the default inline player for HLS (`.m3u/.m3u8`) sources it can't demux.
[...]

PR #1092, opened 2026-06-25
feat(playback): add Ferrite (WASM/WebCodecs) inline player for software HEVC

multiduplikator:
Adds **Ferrite** as a fourth inline web player — a canvas + WebAssembly player ([`ferrite.js`](https://www.npmjs.com/package/ferrite.js), an mpegts.js-style drop-in) that software-decodes **HEVC** where the browser/PWA has no hardware HEVC decoder (and uses WebCodecs when it does). Handles **live and VOD**, rendering to a `<canvas>` with WebAudio.

### Why
On a machine/browser with no hardware HEVC decoder, `<video>`/MSE and the existing inline players can't play HEVC. Ferrite carries its own FFmpeg subset compiled to WASM and decodes it on worker threads, fast enough for live 4K — covering the gap mpegts.js can't (it's bound to MSE/`<video>`).

### What Ferrite brings (beyond decoding HEVC)
Capabilities that matter for broadcast/IPTV content:
- **Broadcast-grade video** — 10/12-bit decode; HDR (PQ/HLG, BT.2020) tone-mapped to SDR; correct per-stream color (BT.601/709/2020 + limited/full range); anamorphic SAR→DAR (1440×1080 4:3-pixel → 16:9); and software **bwdif deinterlacing** for interlaced broadcast.
- **Smooth playback** — an audio-master clock with a PLL-paced present loop, gentle live latency catch-up (≤1.05× toward the edge), and a display-cadence engine that stays smooth at any refresh (50-on-60, 25-on-75…). Under sustained load it degrades gracefully (skip non-reference frames → skip in-loop deblock → halve present rate) to keep audio and motion intact, recovering automatically.
- **Resilience for flaky streams** — reconnect-with-backoff at clean boundaries, an adaptive silence watchdog that re-opens a stalled-but-open socket, WebCodecs decoder self-heal (rebuild → software fallback), and state-machine teardown with no resource creep across play/stop.
- **Robust mid-stream** — resolution changes, cross-codec failover, and sparse parameter sets handled without a reload.
- **Architecture** — split-realm pipeline: a decode worker, a present worker (owns the OffscreenCanvas + WebGL2 + present loop), and a thin main-thread facade (audio + events) — so decode and present never contend.

### Integration
- New `VideoPlayer.Ferrite`, selectable in **Settings → Player** and the workspace command palette. `web-player-view` routes live & VOD to it when selected, and falls back to the default inline player for HLS (`.m3u/.m3u8`) sources it can't demux.
- Depends on **`ferrite.js@^1.3.0`** (zero transitive deps). The engine `.wasm/.mjs` + the two worker chunks are copied from `node_modules` into the served assets by the web build's asset globs and fed to the facade as same-origin `workerUrl`/`presentWorkerUrl`.
- Supported by the bundled engine: MPEG-TS / MP4 / MKV containers (probed by content); H.264 / HEVC / MPEG-2 video; AAC / AC-3 / E-AC-3 / MP2 / MP3 audio.
- On-canvas Material controls (play/mute/volume/fullscreen, a VOD seek bar with resume, a software-tier deinterlace select) + a long-press diagnostic overlay.

### Cross-origin isolation (please read)
Ferrite needs **`crossOriginIsolated`** (COOP `same-origin` + COEP `credentialless`) for `SharedArrayBuffer`. When it's off it **degrades gracefully** — the facade emits an explicit unsupported-codec error that the existing diagnostics surface classifies; other players are unaffected.
- Isolation is **opt-in for dev**: a dedicated `web` serve config (`pnpm nx serve web --configuration=ferrite`) sends the headers; the **default serve options are unchanged**.
- Under Electron `file://` the context is never isolated, so Ferrite is effectively a **web/PWA** feature.
- **Production deployment note:** the hosting layer (nginx/CDN/etc.) must send those COOP/COEP headers for Ferrite to activate.

### Docs & tests
- Documented in `docs/architecture/embedded-inline-playback.md` (+ a README feature bullet).
- Unit tests for the component, controls bar, and event-wiring; i18n keys across all locales; theme-token styling; lint clean. The component session lifecycle is factored into `ferrite-session.ts` to keep files within the repo's length conventions.


multiduplikator:
Addressed all three findings in 52d73d9:

- **Canvas key omits stream params** — `canvasKey` now keys on the effective URL (`url + epgParams`, matching `playChannel`), so an epgParams-only change recreates the canvas instead of re-attaching a transferred element. Added a regression test.
- **Resume handler survives teardown** + **detach mismatch** — `stopFerriteSession` now simplifies to `player.destroy()`, ferrite's single canonical teardown: it terminates both workers (releasing the transferred OffscreenCanvas), reaps the pthread pool, **and clears all event listeners** — so a still-pending VOD resume `MEDIA_INFO` handler can't fire a `seek()` against a torn-down session. Dropped the redundant pause/unload/detachMediaElement.

ui-playback unit tests (138) + lint + web build are green; component is 337 lines.

multiduplikator:
Superseded by #1098 — rebased onto current `master` (this PR had drifted into merge conflict) and updated to the just-released ferrite.js 1.3.2 (audio-path overhaul + the new Dyna audio-dynamics control). All review findings from #1091/#1092 remain fixed. Continuing in #1098.

PR #1102, opened 2026-06-29
feat(epg): rework live EPG panel into a horizontal timeline

4gray:
## Summary

Replaces the vertical EPG list/panel below the player with a single **horizontal timeline ribbon** (`app-epg-timeline`), shared across all four live surfaces — the M3U player, the unified live tab, and the Xtream and Stalker layouts. It preloads a multi-day window and supports scroll, zoom, day navigation, short-programme tiers, and per-state empty/error views. The programme-details dialog is restyled to match and now shows the channel logo.

## Timeline / UI

- **Auto-focus on channel select** — lands on the currently-airing programme instantly (`behavior: 'auto'`, no scroll animation), centred, and deduped by programme-set identity so the 30s tick / zoom / data re-emits never cause flaky re-jumps. The explicit **Now** button still animates.
- **Wide-tier description preview** — long/zoomed cards with a short title show a dimmed description preview underneath. Title and preview clip on **whole lines** (`flex-shrink: 0` + `-webkit-line-clamp`), never mid-line.
- **Dynamic bar subtitle** — shows the now-playing (or archive) programme title instead of a static "Timeline" label; falls back to the source label only when collapsed or with no EPG.
- **State-aware toolbar** — Now + zoom are hidden when there's no ribbon, and the date stepper is hidden when the channel has no EPG on any day (hidden, not disabled — a disabled control would imply it could work).
- **Empty-state** fills and centres within the panel area instead of a fixed `min-height` that overflowed the compact inline panel and pushed the icon/text and action buttons below the visible edge.

## EPG current-programme lookup fixes (electron-backend)

Several distinct bugs caused the **sidebar and timeline to disagree** about whether a channel had a current programme:

- **Timezone-aware "airing now"** — compare with SQLite `datetime()` on both sides instead of a lexical string compare, which broke when stored timestamps carry an offset (`+03:00`) but `now` is UTC (`…Z`).
- **Unscoped fallback** — when a channel resolves but its programmes are tagged with a `source_url` outside the active scope (multi-source imports sharing XMLTV ids), retry once unscoped, mirroring the timeline.
- **Non-ASCII channel names** — match candidates by the raw, case-sensitive key as well as `LOWER()`, since SQLite `LOWER()`/`COLLATE NOCASE` fold only ASCII. Fixes Cyrillic/Greek channels showing "No program information available".
- **Midnight-spanning programmes** — `hasProgramsForDateKey` is now overlap-based, so a film that starts the previous evening and runs past midnight keeps the ribbon on "today" while it airs.

## Refactor

- Extract the duplicated channel-list EPG enrichment (progress % + side-car map) into `channel-list-container/epg-enrichment.util.ts`, shared by the all-channels / groups / favorites / recent views. Also fixes recent's and favorites' previously-unguarded, NaN-prone progress maths.

## Testing

- Unit: **ui-epg 75**, **components 77**, **electron-backend 432** — all green. Lint clean across affected projects; `nx build web` + `nx build electron-backend` succeed.
- New coverage: auto-focus centring/dedup, wide-tier preview gating + whole-line clamps, toolbar visibility gating, overlap-based day membership, `datetime()` + unscoped-fallback batch lookups, and the shared enrichment util.
- Manually verified live in Electron via CDP across M3U channels (ribbon rendering, sidebar/timeline parity for tz/scope/Cyrillic/midnight-spanning cases, empty-state centring, dynamic subtitle, and state-aware controls).


4gray:
Thanks for the reviews — went through all five findings. Verified each against the code; no false positives. Resolved in `23f20a1a`:

| # | Finding | Verdict | Resolution |
|---|---------|---------|------------|
| ① | Codex — auto-focus forces empty-day back to today | ✅ Real bug | `maybeAutoFocus` now skips without clearing `lastFocusKey` and only commits today when today actually has a programme — so navigating from an empty day to a day with data no longer snaps back and traps the user. |
| ② | Codex — unified tab omits `archiveDays` | ✅ Valid | Pass `[archiveDays]="timelineArchiveDays()"` (via the shared `getM3uArchiveDays`), so M3U catch-up "Watch" is gated to the channel's real window instead of the unlimited `0` default. |
| ③ | Greptile — on-air programme swallowed into a group chip | ✅ Valid (edge case) | Grouping now skips the `when: 'now'` block, keeping its highlight + auto-focus target when zoomed out. |
| ④ | Greptile — SQL parameter count → `SQLITE_LIMIT_VARIABLE_NUMBER` | ⚠️ Valid (limit is 32766, not 999) | The candidate query binds each key up to 4× (raw/lower × id/displayName). Chunked to 500 keys/query (~2k params) so it stays well under the limit on large playlists. |
| ⑤ | Greptile — extra EPG-metadata IPC per channel switch for the logo | ⚠️ Minor, deferred | The `switchMap` already cancels in-flight requests on rapid switches and the metadata lookup is a fast indexed query, so the real-world cost is negligible; deferring the icon fetch to dialog-open would add coupling for little gain. Happy to do it if preferred. |

Each fix has regression coverage (auto-focus dedup/guard, now-block grouping, candidate-lookup chunking, archive-days binding). Local: ui-epg 78, electron-backend 433, portal-shared-ui 47; lint + web build green.

4gray:
@greptileai

4gray:
@codex review

The two Codex findings from the previous review (auto-focus forcing empty-day back to today; missing `archiveDays` on the unified live tab) are addressed in `23f20a1a`, along with the two Greptile findings (now-short group-chip guard; candidate-lookup chunking). Please re-review the latest commit.

4gray:
@codex review

Fixed in `a4251d88`: `epgArchiveDays` in the M3U player now uses the shared non-blank `getM3uArchiveDays` helper (skips `tvg-rec=""` and keeps catchup>timeshift>tvg-rec precedence), matching the unified live tab. Added a regression test for the blank-`tvg-rec`-with-real-`timeshift` case. Thanks for the catch.

4gray:
@codex review

4gray:
@greptileai

4gray:
@codex review

4gray:
@greptileai

4gray:
@codex review

4gray:
@greptileai

4gray:
@codex review

4gray:
@greptileai

4gray:
@codex review

4gray:
@greptileai

4gray:
@codex review

4gray:
@greptileai

PR #1104, opened 2026-06-30
Pr/epg fixes

Alanon202:
I’m using IPTVNator with xstream providers and my experience left a lot to be desired so I tried to fix some of that.

I did some work on EPG and catch-up features which seemed broken in several ways, especially when browsing from the favourites tab. This PR tries to fix many problems I encountered. 

While fixing stuff like not being able to load all the EPG data properly from the favourites list, not being able to activate catch-up, I also tried to fix some other things as well.

Some caveats:
- I’m a crappy coder, so the code is probably janky in places, my hope was mainly that this is useful as a baseline for someone that knows better;
- My abilities to iterate on this code are probably going to be limited, but I’ll try my best;
- I was only able to test it out on xstream streams from two providers - further testing is needed for m3u to make sure I didn’t break something;
- I was only able to test it out on Linux, further testing is needed on other platforms;
- Currently all internal players seem to have buffering issues during catch-up and stop after 3 minutes (mpv is fine) - this could be due to my providers or the players themselves, TBD;

### DIFF
--- a/apps/electron-backend/src/app/api/main.preload.ts
+++ b/apps/electron-backend/src/app/api/main.preload.ts
@@ -517,6 +517,14 @@ const electronApi: ElectronBridgeApi = {
         ipcRenderer.invoke('EPG_CHECK_FRESHNESS', { urls, maxAgeHours }),
     searchEpgPrograms: (searchTerm: string, limit?: number) =>
         ipcRenderer.invoke('EPG_DB_SEARCH_PROGRAMS', searchTerm, limit),
+    getEpgMapping: (channelKey: string) =>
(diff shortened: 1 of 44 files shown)


Alanon202:
@greptileai Commit 005d2fa adresses your concerns, please conduct another review:

- add 14-day time window filter to EPG queries (replaces unbounded .limit removal)
- respect preferUploadedEpgOverXtream setting in loadXtreamEpgItems
- extract catchup logic into UnifiedLiveCatchupService (component: 724→571 lines)
- remove duplicate credential-fetching logic (now in shared service)

PR #1105, opened 2026-06-30
Proposal: unified player-controls architecture (docs only, for direction)

larsemig:
> [!NOTE]
>  **This is a direction proposal — docs only, no code.**
>  I'd like your call on whether IPTVnator should move toward this architecture proposal **before** I open implementation PRs. The code already exists. (demonstration video below)

## Where does this come from

For me (macOS user) the embedded MPV player just works best.

### I wanted to fix a "simple" bug
- On macOS, the embedded-MPV in fullscreen, the controls always inset the video resulting in  **black bars**
- The controls were nearly invisible (dark-on-dark).

I thought I can "just" overlay the controls, fix some CSS and I am done.
As you already know: there is a native window composition rabbit hole.

### Rabbit hole ;tldr
- Tried native surface **above** the web UI
- Tried native surface **below** the web UI
- Hours later gave up
- Went back to "docked" and just "bugfix" and UI/UX improvements.
- Fixed the bugs but immediately hated the usability of the player. Especially when opening a dropdown: Either the image turns tiny or disapers completely.
- Dove back into the hole, on principle

3 day later and around 350$ worth of tokens I came up with this proposal:

### Proposal
I eventually committed to the transparent window appraoch with an underlying native window.
The main advantage being that all wiring of the components (overly) can stay (as is) in the web app.

### But why even bother? While I was down there anyway...
- Noticed embedded-MPV was the **only** player with custom controls — every web engine (Video.js, hls.js/HTML5, ArtPlayer) shipped its own vendor skin, so the UX changed depending on which player you landed on.
- That felt wrong, so I extracted **one shared controls layer** every engine reuses:
  - a single controls component instead of N vendor skins
  - an engine-agnostic **contract** (capabilities + state + commands); each engine is just a thin **adapter**
  - identical look, behaviour & keyboard shortcuts independently of chosen player
  - **fully customizable** in appearance *and* functionality — our component, not a vendor skin
  - build a feature once → every player inherits it
- Existing web players are wired onto it **behind a feature flag (default off)** — opt-in, nothing changes for web users until flipped.

## What's in this PR
Docs only:
- `docs/proposals/player-controls-architecture.md` — the proposal: required vs. optional steps, advantages, downsides, and open questions for you.
- `docs/architecture/player-controls-contract.md` — contract, shared component, adapters, flags.
- `docs/architecture/player-controls-refactor.md` — immersive-overlay design + the full decision log.
- `docs/architecture/embedded-mpv-native.md` — native/IPC internals, immersive compositing, native fullscreen.

## Scope, platforms & safety
- **Shared web controls:** all platforms, **flag-gated, default OFF** — zero behaviour change until enabled. Rollback = leave the flag off.
- **Embedded-MPV immersive + native fullscreen:** **macOS only** (transparent-window compositing). Windows/Linux embedded MPV unaffected; gated by the existing `IPTVNATOR_ENABLE_EMBEDDED_MPV_EXPERIMENT` env.
- The ~743-line embedded-MPV component was split into focused, single-purpose files.

## Trade-offs
- On macOS, the transparent window drops `titleBarOverlay` functionality on macOS but since it wasn't used we only lost the possibility to use it in the future. (native traffic lights are still there).
- A bug can actually cause the application(or parts of it) to be transparent exposing the content/desktop below the application. Not sure if we can always catch this with tests.

If you are willing to move in this architectural direction, I can provide these changes though a series of PRs.

This video demonstrates my current implementation (first player is embedded MPV) with the feature flag turned on (all players use the same controls (2x speed)

https://github.com/user-attachments/assets/cc14086a-75b1-47da-8e7c-dc51bd6fd4fb


4gray:
Thanks a lot for this, and especially for going down the native-composition rabbit hole so thoroughly. 

**Quick context:** I just shipped 0.22, so we have a clean baseline to build on and no release pressure hanging over this work.

On the direction:

**ADR-0001** (unified controls contract + adapters) - yes, let's do it. 

**ADR-0002 / 0003** (immersive compositing + native fullscreen) - the video looks great, and I want to try it. The immersive overlay genuinely looks like it solves the airspace problem that the docked strip never could (video shrinking away when a menu opens was always the worst part). I want to see it running on-device before committing, since the transparent: true window is an app-wide setting and the fullscreen freeze/reconciliation logic is the delicate bit - but the concept and the demo are convincing. I understand and accept that this part is macOS-only; Windows/Linux staying on the docked path is fine.

larsemig:
The diff is quite big...
Since you are interested, my next steps will be:
1. Rebase my work on current main,
2. Create multiple PRs (at least split like the ADRs) so it is easier to review.
3. PRs will depend on each other so you can check out the last one and test everything.

Tell me what you want to do with the ADRs (discard or pull in) 
If pulled in, I would transform them into architecture files (codebase standard) in the corresponding PRs.

I cannot test linux or windows thats why I didn't change anything (already had insane scope creep) 
If we have docked players there as well we might also add a flag which can toggle between the player controls. (If everything works, we can maybe delete the old players).
However I will start with macOS only as discussed.

4gray:
Closing this proposal as superseded by the shipped implementation, not rejected.

The ADR-0001 direction landed through #1148/#1149 and the replacement integrations #1193–#1198, with additional shared-controls work in #1199.

The transparent-window and native-fullscreen proposals from ADR-0002/0003 were deliberately superseded by the frame-copy architecture in #1169/#1171/#1175; the corresponding implementation PRs #1150 and #1151 have already been closed.

The authoritative shipped architecture now lives in `docs/architecture/player-controls-contract.md` and `docs/architecture/embedded-mpv-native.md`.

Thank you for proposing and implementing the foundation that led to the current architecture.

PR #1106, opened 2026-06-30
Pr/mapping

Alanon202:
**Note**: The first 4 commits are from the epg-fixes PR. Only the last 3 are new here.

This set of fixes has a lot of quality of life improvements for my use case at least. 

Mainly it adds a context menu to and accompanying GUI with the option to manually map channels to user-provided EPG data, along with improvements about how the data is loaded and shown (for some reason beforehand this was much slower and very inconsistent).

Another pet peeve of mine I was able to resolve was the favourites list which did not remember my custom channel order.

The same caveats apply:
- I’m a crappy coder, so the code is probably janky in places, my hope was mainly that this is useful as a baseline for someone that knows better;
- My abilities to iterate on this code are probably going to be limited, but I’ll try my best;
- I was only able to test it out on xstream streams from two providers - further testing is needed for m3u to make sure I didn’t break something;
- I was only able to test it out on Linux, further testing is needed on other platforms;
- Currently the EPG popup menu is a bit basic, but works;

### DIFF
--- a/apps/electron-backend/src/app/api/main.preload.ts
+++ b/apps/electron-backend/src/app/api/main.preload.ts
@@ -517,6 +517,14 @@ const electronApi: ElectronBridgeApi = {
         ipcRenderer.invoke('EPG_CHECK_FRESHNESS', { urls, maxAgeHours }),
     searchEpgPrograms: (searchTerm: string, limit?: number) =>
         ipcRenderer.invoke('EPG_DB_SEARCH_PROGRAMS', searchTerm, limit),
+    getEpgMapping: (channelKey: string) =>
+        ipcRenderer.invoke('EPG_MAPPING_GET', { channelKey }),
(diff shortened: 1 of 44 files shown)


Alanon202:
@greptileai commit 495c344 resolves your issues, please rescan:

- remove dead this.search$ line in dialog constructor
- check save/remove result before closing mapping dialog
- escape LIKE wildcards (% and _) in EPG channel search
- remove double mapping lookup (epg.events IPC handler no longer resolves mappings since epg-query.service already does)

Alanon202:
@greptileai review commit https://github.com/4gray/iptvnator/commit/7a23ba7a4c09f5f00f2161b5f396f0883275f710 for the last remaining issue you flagged.

Alanon202:
@greptileai here you go, commit 1eeea1d for the last remark.

PR #1098, opened 2026-06-28
feat(playback): add Ferrite (WASM/WebCodecs) inline player for software HEVC

multiduplikator:
## Ferrite (WASM/WebCodecs) inline player for software HEVC

Adds **Ferrite** as a new inline web player — a `<canvas>` + WebAssembly software decoder with an automatic **WebCodecs** hardware tier — so the web/PWA build can play **HEVC where the browser has no hardware decoder** (and H.264/MPEG-2, 8/10-bit, interlaced). Requires a cross-origin-isolated context (COOP/COEP); falls back to the default inline player for HLS (`.m3u`/`.m3u8`).

> This supersedes #1092 — rebased onto current `master` (that PR had drifted into conflict) and updated to the just-released **ferrite.js 1.3.2**. All earlier greptile findings from #1091/#1092 remain fixed (canvas key keyed on `url + epgParams`; teardown via the facade's canonical `destroy()` which releases the transferred canvas and clears listeners; no resume listener survives teardown).

### What's in it
- **`ferrite-player`** component + session/event wiring (live + VOD), IPTVnator-owned Material controls and a long-press diagnostic overlay. Teardown goes through `destroy()` only.
- **Routing/settings/command-palette/i18n** registration; HLS falls back to the default inline player.
- **Engine + worker asset bundling** (`ferrite.{mjs,wasm}`, `worker.js`, `present-worker.js`) into `apps/web` assets, plus an opt-in COOP/COEP isolated dev-serve configuration (`nx serve web --configuration=ferrite`).

### New in this revision (ferrite.js 1.3.2)
- Bumped `ferrite.js` `^1.3.0` → `^1.3.2`. The 1.3.2 audio-path overhaul (clean 5.1→stereo downmix + device-rate resample, loudness leveling), all-codec anamorphic SAR, and correct HEVC sizing on browsers with bogus `VideoFrame` dims (Edge) are **internal to ferrite.js** and apply automatically via the re-bundled engine.
- **New "Dyna" audio-dynamics control** (Line / RF / Night) in the player controls, driving the facade's `setDrc()`. Always shown (audio is always decoded on both tiers); resets to Line per stream like the deint mode. *Night* is a built-in compressor that tames loud peaks on any codec — useful for late-night listening.

Deliberately left out (no public ferrite.js API / not appropriate as app settings): the fast non-conformant decode toggle and a persisted default-deinterlace setting.

### Validation
- `pnpm nx lint ui-playback` ✓
[...]

PR #1111, opened 2026-07-01
feat(playback): add Ferrite (WASM/WebCodecs) inline player for software HEVC

multiduplikator:
## What

Adds **Ferrite** — a `<canvas>` + WebAssembly/WebCodecs player ([`ferrite.js`](https://www.npmjs.com/package/ferrite.js)) — as a fourth inline web/PWA player, so **HEVC plays where the browser has no hardware decoder** (the current gap: `<video>` + hls.js/mpegts.js can't software-decode HEVC). It has an automatic WebCodecs hardware tier and falls back to the default inline player for HLS (`.m3u`/`.m3u8`).

Requires a cross-origin-isolated context (COOP/COEP) for `SharedArrayBuffer`; when isolation is off it degrades gracefully with an explicit unsupported banner (no black screen).

## How it's wired

- `FerritePlayerComponent` (same component I/O as `HtmlVideoPlayerComponent`, rendering to a canvas) with the separable concerns split out — `ferrite-session`, `ferrite-event-wiring`, `ferrite-controls`, `ferrite-debug-overlay`, `overlay-interactions`.
- Slotted into `WebPlayerViewComponent`'s `selectedPlayer()` ladder as `VideoPlayer.Ferrite`; teardown always via the facade's canonical `destroy()` (releases the transferred `OffscreenCanvas` + clears all listeners).
- IPTVnator-owned Material controls (play/mute/volume/fullscreen, VOD seek bar, software-tier deinterlace select, and a "Dyna" audio-dynamics select) + a long-press diagnostic overlay (surfaces `isolated: NO` on a devtools-less PWA).
- Settings + command-palette registration, i18n (23 locales), the diagnostics model, and asset bundling with an **opt-in** `nx serve web --configuration=ferrite` that adds the COOP/COEP headers for local testing. No change to any existing player path.

## ferrite.js 1.3.4

Bumps the engine to `1.3.4`, which moves to a **four-worker topology** (demux · video-decode · audio-decode · present). The four self-contained worker chunks are copied into the served assets and fed to the facade via its `{worker,presentWorker,audioWorker,demuxWorker}Url` overrides — the esbuild builder does not rewrite a dependency's internal `new Worker(new URL('./x.js', import.meta.url))`, so the explicit same-origin URLs are required for the isolated context. Audio now runs off the main thread (an AudioWorklet-driven master clock), and A/V sync uses an mpv-faithful display-sync corrector (tight lip-sync after pause/resume).

## Review history

Supersedes #1098 (and the earlier #1092 / #1091). This is a clean, rebased, single-commit re-submission so the review bot gets a fresh pass. All prior findings are incorporated: no hardcoded `BACKEND_URL`; `canvasKey` keys on the effective URL (`url + epgParams`); `formatPlayer` has the `'ferrite'` arm; the `ferrite` dev-serve uses the development build. On the earlier soft note about `ferrite-player.component.ts` length — the separable concerns are already extracted into the five sibling files above, and what remains is the cohesive lifecycle core (it passes lint; there is no `max-lines` rule).

## Testing

- `nx lint ui-playback` clean; `nx test ui-playback` green (140 tests, incl. the ferrite specs — the component spec locks the four-worker wiring).
- Local smoke test in a cross-origin-isolated dev-serve: live + VOD HEVC playback, pause/resume lip-sync, channel-zap teardown (no orphaned workers), and the HLS → default-player fallback.


multiduplikator:
Thanks for the review! Addressed the two P2 findings:

- **`allowedHosts: true` + `host: 0.0.0.0` in the `ferrite` dev-serve config** — removed both from the committed config, so it now defaults to localhost. The COOP/COEP headers (the actual point of the config) are unchanged, so cross-origin-isolated testing still works; network/on-device testing stays available as an explicit opt-in (`nx serve web --configuration=ferrite --host 0.0.0.0 --allowed-hosts all`). No wide-open default ships in the repo.
- **`ferrite-player.component.ts` at 346 lines (>300 soft guideline)** — left as-is: it's under the 350 hard limit and passes lint, and the separable concerns are already extracted into the five sibling files (`ferrite-session`, `ferrite-event-wiring`, `ferrite-controls`, `ferrite-debug-overlay`, `overlay-interactions`). What remains is the cohesive component lifecycle; splitting it further would push Angular-bound player/signal handlers into context-passing free functions, which reads worse. Happy to break it up differently if you'd prefer a specific split.

multiduplikator:
Superseded by #1119 — the same integration rebased onto current master and bumped to the newly-published ferrite.js@1.3.5 (mpv-faithful live pause/resume, WebCodecs resume lookahead, engine on emscripten 6.0.2). Closing in favor of #1119.

PR #1142, opened 2026-07-06
feat(portals): back button on the in-portal search page

4gray:
## Problem

From an actor page, clicking a "grey" (unmatched) film opens the portal search prefilled with `?q=`. But the search page had **no way back** — to return to the actor's filmography you had to re-navigate through the sidebar and search for the actor again. (Reported by the user.)

## Fix

`SearchLayoutComponent` (shared search shell) gains a `showBackButton` input + `backClick` output. Both search hosts wire it to `Location.back()`:
- **Xtream** `SearchResultsComponent` — shown when `isWorkspaceLayout && !isGlobalSearch` (in-portal search only).
- **Stalker** `StalkerSearchComponent` — shown on the in-portal search (its only mode).

Excluded on purpose:
- **Global search** (`/workspace/search`, sidebar destination) — top-level, no back needed.
- **Dialog mode** — already has its own close button.

`location.back()` unwinds the full history, so the whole chain works: search → actor page → movie detail → cast → … Each step returns where the user came from. It also naturally covers the toolbar "Search in this playlist" flow (back returns to the content you were viewing).

## Why always-on for in-portal search (vs a "came-from-actor" flag)

The in-portal search is always a *nested* view reached by an action; a back arrow is natural there and matches the detail-page pattern. It's simpler and more robust than threading a navigation-state flag (no timing issues with component reuse / query-param changes), and it helps the toolbar-search flow too.

## Validation

- `nx test portal-shared-ui portal-xtream-feature portal-stalker-feature` — green, incl. new search-layout spec (conditional button + `backClick` emit)
- `nx build web` ✅, lint 0 errors ✅
- Existing search E2E untouched (they drive the top toolbar search field, not the layout header)
- Docs: CLAUDE.md actor-pages note updated


4gray:
@greptileai @codex review — addressed the two P2 comments: showBackButton now also guards against dialog mode (`!this.dialogRef`), and the fragile negative margin was replaced with an RTL-safe `.header-title-group` flex wrapper (gap). Please re-review.

PR #1143, opened 2026-07-06
fix(favorites): persist custom drag-and-drop order for Xtream favorites

### LINKED ISSUE #1137: Global Favorites custom order is not persisted after drag & drop
**Describe the bug**

When the Global Favorites list is sorted using **Custom Order**, manually reordering channels via drag & drop is not persisted.

The new order is visible immediately after dragging, but as soon as I switch to another section/tab and return to Favorites, the list reverts to its previous order.

**To Reproduce**

1. Open **Global Favorites**.
2. Set the sorting mode to **Custom Order**.
3. Drag one or more favorite channels to a different position.
4. Switch to another section (for example Live TV, Movies, or Series).
5. Return to **Global Favorites**.

**Expected behavior**

The manually arranged order should be saved and remain unchanged after navigating away from the Favorites view.

**Screenshots**

Image

- PWA or Electron/Tauri application: **Electron (.deb)**
- OS: **Linux Mint 22.3**
- Browser: N/A
- Version: **v0.22.0**
- Feature: **Global Favorites**

**Additional context**

- This issue was also present in previous versions.
- The problem is consistently reproducible on Linux Mint 22.3.

4gray:
## Summary

Fixes the custom favorites order not persisting after drag & drop for the **"This playlist"** Xtream scope (#1137).

### Root cause

Prepared-statement writes were dispatched with drizzle's `.execute()`, which on the `better-sqlite3` driver **returns a promise and defers the write to a microtask**. Both writers run their statements inside a *synchronous* `db.transaction(() => …)` callback that can't `await`, so the transaction commits before the promise settles — the write is a **silent no-op** (no error, no rows changed).

This left `favorites.position` untouched, so on reload the list fell back to `added_at DESC` and appeared to "revert".

Why only **"This playlist"** was affected: that scope orders purely by the DB `position` column. The **"All playlists"** scope also persists an order to the `appState` `global-favorites-channel-order-v1` key and re-applies it on read, which masked the broken DB write.

The mocked operations specs never caught it — a jest mock records an `.execute()` call the same as a `.run()` call.

### Fix

- `reorderGlobalFavorites` and `removeRecentItemsBatch`: switch the prepared statement from `.execute(...)` to synchronous `.run(...)`.
- `removeRecentItemsBatch` had the same latent bug — batch "clear recent items" was silently doing nothing; fixed too.

### Tests

- Extended the operations test mock to model `.prepare().run()`.
- New regression test for `reorderGlobalFavorites` and rewrote the `removeRecentItemsBatch` test (it previously asserted the buggy `.execute()`). Both now assert `.run()` is used and `.execute()` is not — verified they **fail** on the old code and **pass** with the fix.
- `pnpm nx test electron-backend` → all pass; `pnpm nx lint electron-backend` → clean.

### Manual verification

Verified over CDP against a live Electron instance in "This playlist" Xtream scope:

| Step | Result |
|---|---|
| Reorder (move last to top) | DB positions written `0..N` ✅ |
| Navigate away → back | Order reloaded from DB, persisted ✅ |
| Full page reload | Order still persisted ✅ |

### Docs

Added a "Prepared-statement writes must use `.run()`" gotcha section to `docs/architecture/sqlite-db-worker.md`.

Fixes #1137

### DIFF
--- a/apps/electron-backend/src/app/api/main.preload.spec-data.ts
+++ b/apps/electron-backend/src/app/api/main.preload.spec-data.ts
@@ -34,7 +34,9 @@ const streams = [{ stream_id: 42, name: 'Channel' }];
 const favorites = [{ contentId: 1, playlistId }];
 const recentlyViewed = [{ contentId: 2, playlistId }];
 const categoryIds = [10, 11];
-const reorderUpdates = [{ content_id: 12, position: 1 }];
+const reorderUpdates = [
+    { content_id: 12, playlist_id: 'playlist-1', position: 1 },
+];
 const recentItemsBatch = [{ contentId: 13, playlistId }];
 const playbackData = {
     contentXtreamId: 42,
--- a/apps/electron-backend/src/app/api/main.preload.ts
+++ b/apps/electron-backend/src/app/api/main.preload.ts
@@ -710,7 +710,7 @@ const electronApi: ElectronBridgeApi = {
     dbGetAllGlobalFavorites: () =>
         ipcRenderer.invoke('DB_GET_ALL_GLOBAL_FAVORITES'),
     dbReorderGlobalFavorites: (
-        updates: { content_id: number; position: number }[]
+        updates: { content_id: number; playlist_id: string; position: number }[]
     ) => ipcRenderer.invoke('DB_REORDER_GLOBAL_FAVORITES', updates),
     // Recently viewed (playlist-specific)
     dbGetRecentItems: (playlistId: string) =>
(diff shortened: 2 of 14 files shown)


4gray:
@greptileai @codex review — fixed the P1: the favorites reorder now threads `playlist_id` through the whole path and scopes the prepared UPDATE by `(contentId, playlistId)` (matching the favorites composite unique index), so a same-`content_id` favorite in another playlist can no longer be clobbered. Regression coverage added in favorites.operations.spec. Please re-review.

4gray:
@greptileai

4gray:
@greptileai @codex review — addressed the remaining findings in 7ba58592: the workspace global-favorites reorder path (`GlobalFavoritesService.reorder()`) now includes `playlist_id` in every update, so the payload matches the playlist-scoped `(contentId, playlistId)` UPDATE predicate, and the prepared-statement example in `docs/architecture/sqlite-db-worker.md` is now playlist-scoped as well. Added a regression spec that fails on the old payload shape. Please re-review.

PR #1162, opened 2026-07-10
feat: add strip country prefix setting

Alanon202:
Adds a settings toggle to strip country/group prefixes from channel names (e.g. "US | CNN" → "CNN", "UK - BBC One" → "BBC One"). The strip is applied in the channel list items, grid list view, EPG timeline header, external player footer dock, and the inline player header. The prefix detection handles common separators (|, -) and finds the earliest match so multi-segment names like "ES - A3 - Sports" correctly become "A3 - Sports".


Alanon202:
@greptileai Here you go, please retest!

4gray:
@greptileai Please re-review — the follow-up commit narrows the strip heuristic (pipes always, dash/colon only for short uppercase tags), scopes stripping to live content only, covers the previously missed M3U player / radio / dashboard surfaces, adds i18n keys for all 18 locales, and adds unit + regression tests.

4gray:
@Alanon202 Thanks a lot for the contribution — nice idea and a clean starting point! 🎉

Before merging I pushed a follow-up on top of your branch, short summary of what changed:

- **Narrower heuristic**: pipes always strip; `-` / `:` separators only strip when the prefix looks like a short tag (`UK`, `USA`, `4K`), so names like "Sky - Sports F1" or "Mission: Impossible - Fallout" stay intact. If stripping would leave an empty name, the original is kept.
- **Live-only scope**: VOD/series titles are never stripped (grid type, `isLive`, external session `contentInfo`, dashboard card `contentType`).
- **More surfaces**: M3U player EPG timeline header, M3U inline player title, radio player, and dashboard live rails now honour the setting too.
- **i18n**: the settings toggle is translated into all 18 locales.
- **Tests**: unit spec for the utility + regression specs across the call sites (patch coverage 92%).

Merging now — the toggle ships in the next release. Thanks again! 🙏

Alanon202:
Hey @4gray thanks a lot for the kind words and patience.
I'm glad I could contribute in some small way. 😊

Issue #1138, opened 2026-07-05
[BUG] Past EPG pgms not displayed when accessing a channel thru Favorites


WolfganP:
**Describe the bug**
When accessing a channel with EPG info, the programs shown in the (new vertical) right panel, is different if selected thru the Live TV/Category tab or if directly from the Favorites tab.
Former shows all past and future EPG program information, later only shows Now and future programs only

**To Reproduce**
Locate a live channel with EPG info available, 
favorite it; 
compare the EPG shown accessing it thru the Live TV or Favorite tab

**Expected behavior**
Same EPG information shown and selectable no matter how the live channel is accessed. 

**Desktop (please complete the following information):**

-   Electron/Tauri x86/64 appimage application 
-   OS:Linux 13 Trixie, XFCE
-   Version 0.22
-   Feature: Xtream Code live TV channels

WolfganP:
Update: just tried devel version after https://github.com/4gray/iptvnator/commit/f856226dc360a1a02591bc9d87ece48f6b97ed4b and the same EPG information is shown (past and future) on both Live and Fav tabs

The only remaining bug is that for a channel with catch up availability, past programs can only be accessed via the Live TV tab (same channel in Favorites shows " Schedule only. This provider exposes programme history but no catch-up")

WolfganP:
@4gray @Alanon202 Seems to be working OK now after https://github.com/4gray/iptvnator/pull/1166
Thanks a lot for the prompt implementation!!!

PR #1166, opened 2026-07-10
feat: add catch-up functionality from global favourites & catch-up of currently live programme

Alanon202:
Extends the existing catch-up infrastructure to the global favorites list and the EPG timeline. Adds tv_archive/tv_archive_duration through the data pipeline (DB queries → data service → UI component) so the timeline can detect when a provider has timeshift/archive enabled.

The Watch from start button now appears on the current program in the timeline, and catch-up URL resolution works for both M3U and Xtream portals in the favorites context. Includes serverTimezone support for correct Xtream timeshift URL generation.


Alanon202:
@greptileai, I believe I resolved the issues you flagged, can you review?

4gray:
Thanks @Alanon202 for taking this on — the favorites catch-up flow and the start-over idea are exactly the right direction! 🙏 I pushed four follow-up commits on top of your branch to finish the remaining pieces of #1138:

- **`tv_archive_duration` units** (12105c31): the unified tab divided the value by 24 as if it were hours, collapsing e.g. a 5-day archive window to 1 day. The Live TV tab (`live-stream-layout.controlledArchiveDays`) treats it as days, so the unified tab now passes it through unchanged.
- **Global favorites + recent tabs** (edba62f2): `selectGlobalFavoriteRows`, `getRecentlyViewed` and `getRecentItems` still dropped `tv_archive`/`tv_archive_duration`, so `/workspace/global-favorites`, the "All playlists" scope and all Recent surfaces kept showing "Schedule only". Both columns are now selected and mapped onto `UnifiedCollectionItem`.
- **Spec alignment** (bebd1316): `epg-database.spec` and `epg-archive.util.spec` still asserted the old behavior (full-source EPG wipe, no catch-up on the on-now block) and were failing against your change — they now encode the intended contract. Also removed two unused leftovers in `EpgDatabase`.
- **Regression tests + docs** (f439bce4 and within the fix commits): new portal catch-up specs (days pass-through, gate, epoch-seconds URL resolution, failure feedback) plus projection/mapping regressions, and doc updates in `docs/architecture/`.

Validation: `nx affected -t test` vs master is green across all 28 affected projects; lint passes on every touched project.

WolfganP:
@Alanon202 @4gray thx for the extra functionality. I'll make sure I test it and report back on https://github.com/4gray/iptvnator/issues/1138

PR #1170, opened 2026-07-11
fix(xtream): send player-style User-Agent to avoid WAF challenge

sh3rencr:
## Summary
- Fixes "Test Connection" always failing with "Could not connect to the portal" for Xtream Codes portals that sit behind a WAF (e.g. Cloudflare) configured to challenge generic browser-looking User-Agents while allowlisting known IPTV player clients.
- Root cause: `apps/electron-backend/src/app/events/xtream.events.ts` sent a hardcoded, truncated browser-style `User-Agent` (`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36`) on the `XTREAM_REQUEST` and `XTREAM_PROBE_URL` IPC handlers. Some panels' WAF served a challenge page (HTTP 403 HTML) to that signature instead of the real `player_api.php` JSON response, even though the exact same portal worked fine from curl (with a player-style UA) and Safari.
- Replaced the three inline occurrences with a shared `XTREAM_CLIENT_USER_AGENT` constant using a VLC-style signature, verified against a live affected panel (403/HTML → 200/JSON).
- No changes to TLS handling, the SSRF/private-network guard, or CORS — this was purely a User-Agent/WAF fingerprinting issue, confirmed by reproducing the request from the same Node runtime bundled with this project's Electron build.

## Test plan
- [x] `pnpm nx lint electron-backend`
- [x] `pnpm nx test electron-backend` (48 suites / 513 tests, including 2 new regression tests asserting the User-Agent is no longer the old truncated browser string on both `XTREAM_REQUEST` and `XTREAM_PROBE_URL`)
- [x] Manually verified against a live Xtream panel previously reproducing the bug: `Test Connection` now reports "Connection successful! Portal is active." and the playlist adds successfully

### DIFF
--- a/apps/electron-backend/src/app/events/xtream.events.spec.ts
+++ b/apps/electron-backend/src/app/events/xtream.events.spec.ts
@@ -86,6 +86,45 @@ describe('XtreamEvents session cancellation', () => {
         expect(requestedUrl.searchParams.get('username')).toBe('user');
     });
 
+    it('sends a player-style User-Agent instead of a truncated browser string', async () => {
+        // Regression: some Xtream panels sit behind a WAF (e.g. Cloudflare)
+        // that challenges generic/incomplete browser User-Agents but
+        // allowlists known IPTV player clients. The previous hardcoded
+        // 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36'
+        // string was rejected by such panels even though curl/Safari worked.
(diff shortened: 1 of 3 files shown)


4gray:
Thanks a lot for the contribution and the thorough investigation, @sh3rencr! 🙏 Great catch on the truncated browser User-Agent triggering WAF challenges — the shared constant, regression tests, and docs update are much appreciated. Merging!

PR #1173, opened 2026-07-11
feat(epg): manual EPG mapping for Stalker portals

4gray:
> [!IMPORTANT]
> **Stacked on #1165** — this branch includes all commits from `pr/epg-mapping` and must be merged (or rebased) **after** it. Only the last two commits (`3e99b2ff`, `3d15c685`) are new; the diff will collapse to them once #1165 lands.

Extends the manual EPG-to-channel mapping feature to Stalker portals — the last portal type without it.

## What's included

- **Playlist-scoped keys**: `buildStalkerEpgMappingKey` → `stalker:{playlistId}:{channelId}` in the shared key util, mirroring the Xtream scheme (stalker channel ids are portal-local).
- **ITV sidebar**: right-click any channel → "Map EPG channel" (same dialog as M3U/Xtream). After the dialog closes *with a change* (save or removal — detected by comparing the mapping before/after), the bulk EPG cache is rebuilt so the active panel and row previews update immediately; plain cancel does nothing.
- **Store**: `withStalkerEpg().applyMappedItvEpg(channelIds)` batch-resolves mappings (one `getEpgMappingsBatch` IPC per new id set, deduped per playlist session) and overlays uploaded-XMLTV programs onto `bulkItvEpgByChannel` — the active EPG panel, timeline and row previews pick them up with zero template changes. Overrides survive `ensureBulkItvEpg` reloads.
- **Collections**: `stream-resolver` resolves the stalker key in `loadStalkerEpgItems` (detail) and via the batched prefetch in `loadStalkerEpgBatch` (previews). Stalker items consult *only* the scoped key — their `tvgId` mirrors the raw provider id, so bare candidates could only false-match unrelated M3U mappings.
- **Global favorites**: stalker branch for the Map-EPG entry (item id from the `stalker::{playlistId}::{id}` uid) + a new `epgMappingChanged` output so the unified live tab reloads previews after a change.
- **Docs**: manual-mapping section in `docs/architecture/stalker-epg.md`; CLAUDE.md EPG bullet now documents the whole mapping feature.
- **Hardening from adversarial review**: staleness guard in `applyMappedItvEpg` (an in-flight call can no longer write portal A's EPG into portal B's state after a playlist switch — the store is a root singleton); checked-id bookkeeping happens only after successful lookups so transient IPC failures retry; post-dialog refresh uses the unfiltered channel list.

## Tests

- `applyMappedItvEpg` suite (overlay, per-session dedupe, override survival across bulk reloads, capability guard)
- stalker preview mapping in the stream-resolver spec (batched lookup, portal fetch skipped)
- key-builder specs (scoping, no cross-portal/cross-source collisions)
- all affected projects green: portal-stalker-data-access, portal-stalker-feature, portal-shared-data-access, portal-shared-ui, shared-interfaces; `nx build web` clean

Everything is gated behind `supportsEpgMapping`, so the PWA never sees the menu entry. As with Xtream, the mapping targets uploaded XMLTV — Stalker users without an XMLTV source are unaffected.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -800,6 +800,7 @@ engine` (restart required) or
 - XMLTV format support
 - Background parsing in worker thread
 - Stored in database for quick lookup
+- Manual EPG mapping (Electron only): right-click a channel in any list (M3U views, Xtream portal list, Stalker ITV sidebar, global favorites) → "Map EPG channel" attaches it to an uploaded-XMLTV channel; stored in `epg_channel_mappings` keyed by the M3U lookup key or a playlist-scoped portal key (`xtream:{playlistId}:{id}` / `stalker:{playlistId}:{id}`, helpers in `libs/shared/interfaces/src/lib/epg-mapping-key.util.ts`); resolved on every EPG path (single + batch IPC lookups, portal detail views, preview queues); dialog: `libs/ui/components/src/lib/channel-list-container/epg-mapping-dialog/`
 
 **TMDB Metadata Enrichment** (opt-in):
 
--- a/docs/architecture/stalker-epg.md
+++ b/docs/architecture/stalker-epg.md
@@ -263,6 +263,29 @@ advantage of the richer bulk API when it is available. Row previews do not
 fallback to per-channel requests in this mode; they remain empty until bulk EPG
 is available.
 
+## Manual EPG Mapping
+
+Stalker channels carry no XMLTV identifier, so when the portal's own EPG is
+missing or wrong the only uploaded-EPG entry point is a **manual mapping**:
+right-click a channel in the ITV sidebar (or in global favorites) and pick
+"Map EPG channel" to attach it to a channel from an uploaded XMLTV guide.
+
+- Mappings are stored in the shared `epg_channel_mappings` table under the
+  playlist-scoped key `stalker:{playlistId}:{channelId}`
+  (`buildStalkerEpgMappingKey` in
+  `libs/shared/interfaces/src/lib/epg-mapping-key.util.ts`).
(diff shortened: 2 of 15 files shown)


4gray:
@greptileai @codex

4gray:
### Rebased onto master (conflicts resolved)

#1165 was squash-merged, so this branch carried 6 duplicate commits that conflicted with the merged code. I rebased onto current `master` and the diff is now exactly the **2 Stalker commits** (`feat: manual EPG mapping for Stalker portals` + the adversarial-review hardening). No schema conflict with #1213 — the Stalker delta doesn't touch `schema.ts`. Build / lint / affected unit suites (262 tests) all green.

### On the earlier bot comments

I verified all four against current `master`. All of them point at **already-merged #1165 code**, not at this PR's delta — after the rebase they fall outside the diff:

- **EPG program dedup key omits `source_url`** (Greptile P1 + Codex P2, `epg-database.ts`) — real but minor data-fidelity issue; the now-playing/timeline paths have unscoped fallbacks so the guide still renders. Introduced by #1165.
- **Xtream `getMapping` fallback LIMIT-5 unordered** (Greptile P1, `epg-query.service.ts`) — real minor edge; a manual mapping saved under a later shared-`epg_channel_id` stream can be missed. Introduced by #1165.
- **Xtream mapping dialog has no `afterClosed` refresh** (Codex P2, `portal-channels-list.component.ts`) — real consistency gap vs. the Stalker path added here; the Xtream guide stays stale until TTL/reselection. Introduced by #1165.

None block this PR. The three real ones are worth a **separate follow-up against `master`** (touching core #1165 EPG storage/query code), not an expansion of this Stalker PR.

@greptileai please re-review the rebased diff.

PR #1176, opened 2026-07-12
fix(electron-backend): make test suite and lint host-agnostic on Windows checkouts

4gray:
## What

Makes `nx test electron-backend` and `nx lint electron-backend` pass on Windows checkouts (`core.autocrlf=true`) without changing any behavior on Linux. Baseline on Windows was 13 failing tests across 4 suites plus 5 lint errors; both are now zero on both hosts.

## Why the lint difference existed (reviewer note)

The lint target ran `eslint apps/electron-backend/**/*.ts` with an **unquoted glob**. On Linux the POSIX shell expands `**` shallowly before ESLint runs, so CI only ever linted a small subset of files. On Windows, cmd.exe passes the literal pattern through and ESLint expands it recursively — the full tree. The 5 "Windows-only" errors were real violations in files Linux CI never linted. This PR quotes the glob (same form `tools/packaging` already uses), so both hosts lint the identical full file set, and fixes the 5 surfaced errors so the widened Linux coverage stays green. The convention is now documented in CLAUDE.md's Linting section. (`apps/web-backend` and `libs/shared/database` have the same unquoted pattern; both verified error-free on a full-tree lint, left for a follow-up.)

## Changes

**Test host-agnosticism**
- `embedded-mpv-native-source.spec.ts`: all source reads go through a `readSource()` helper that normalizes `\r\n` → `\n`, so multi-line `toContain` assertions (and the negative ones) work on CRLF checkouts
- `worker-runtime-paths.spec.ts`: expected candidate paths built with `path.join` via helpers mirroring the implementation; `appPath` inputs stay POSIX literals since win32 `path.dirname` preserves separators, keeping the root-dedupe assertions valid
- `external-player-launch-context.ts`: the two darwin-only path builders (`.app` bundle executable, Homebrew Caskroom scan) use `path.posix.join` — identical output on macOS, correct when tests simulate `platform: 'darwin'` on a win32 host
- `app.spec.ts`: packaged-navigation fixtures built with `path.resolve` + `pathToFileURL` (`file:///tmp/...` has no drive letter and is not a valid win32 file URL, so `fileURLToPath` threw); on Linux the produced fixtures are byte-identical to the old literals

**Lint fixes (full-tree errors)**
- `prefer-const`: declare-then-assign-once → `const` in `epg-worker.service.ts` and `database.worker-connection.ts`
- `no-unsafe-finally`: `finally { return ... }` → catch-swallow + unconditional return with identical semantics in `external-player-session-registry.closeSession` and `embedded-mpv-native.service.disposeSession`; two new regression tests pin the preserved behavior (a rejecting closer / throwing native dispose still yields a closed session and registry cleanup)
- `no-control-regex`: justified disable on the recording filename sanitizer (the control-char range is intentional)
- removed 3 stale `no-require-imports` disable directives ESLint reported as unused

## Validation

- `pnpm nx test electron-backend` (Windows): **515/515 tests, 48/48 suites pass** (baseline: 13 failures)
- `pnpm nx lint electron-backend` (Windows): **0 errors, 45 warnings** (baseline: 5 errors, 48 warnings; warnings don't fail the target)
- Linux safety: CRLF normalization is a no-op on LF checkouts; `path.join`/`path.posix.join` reproduce the previous literal expectations on POSIX; the quoted glob widens Linux lint coverage to the file set already verified error-free on Windows


4gray:
@codex review

4gray:
Rebased onto current master (`518964c5`) — the branch was based on `f856226d` from July 12, so it had drifted into a conflict.

Only two files actually conflicted:

**`apps/electron-backend/project.json`** — this PR quoted the glob back when the target had a single one; master has since added `inputs` and two more (already-quoted) globs. Resolved to master's version with the first glob quoted:

```
eslint "apps/electron-backend/**/*.ts" "…frame-copy-runtime.ts" "…frame-copy-runtime/**/*.ts"
```

**`embedded-mpv-native-source.spec.ts`** — this PR introduced the `readSource()` helper (CRLF normalization); master independently added an `electronBuilderConfig` read in the old `readFileSync(path.resolve(...))` style. Resolved by routing the new read through `readSource()` too, and dropping master's now-duplicate `buildScriptSource` / `buildAndMakeWorkflowSource` (this PR had already converted both).

The other 11 files merged cleanly.

### Verification on the merged result

The glob fix is now live locally, so lint finally covers the project:

- **`pnpm nx lint electron-backend`: 245 of 245 files, 0 errors** (57 warnings, all pre-existing). Before the quoting fix `/bin/sh` — which has no `globstar` — collapsed `apps/electron-backend/**/*.ts` to `src/main.ts`, so the target linted **1** file and still reported success.
- `pnpm nx test electron-backend`: **1047 passed, 103 suites**.
- `tsconfig.app.json` and `tsconfig.spec.json` both typecheck clean.

The `max-lines` violations this quoting exposes were split out separately in #1278 and #1288, both already merged — so this PR now lands green rather than surfacing them.

One heads-up: `worker-performance-capture.concurrency.spec.ts` › *"reliably observes a real 20ms event-loop block after condition-based arming"* is flaky on master independently of this PR (it asserts a real measured event-loop block). It failed 4/4 full-suite runs on a clean master checkout here. It passed in the run above, but it may still redden CI at random.

4gray:
Added the `no-release-note` label — the Release note gate was failing because this PR touches `apps/**` runtime files with no `.changes/` note.

No user-visible behavior changes here: the only runtime edits are `path.join` → `path.posix.join` in `external-player-launch-context.ts`, and on darwin those two are the same function. The macOS `.app`-bundle and Homebrew Caskroom paths behave identically in production; the change only matters when a darwin platform is simulated on a win32 host in tests. Everything else is a lint fix or test infrastructure.

Revert the label and add a `.changes/` note instead if you read it differently.

4gray:
@greptileai 
@codex review

4gray:
Fixed the Codex finding — mirrored the quoted-glob convention into `AGENTS.md` (6febf934).

The finding is valid: `AGENTS.md` already mirrors the *neighbouring* paragraph in the same CLAUDE.md section (the `max-lines` / baseline rule, AGENTS.md L19), and its own Documentation-After-Changes section requires coding conventions to stay in sync across both files. So one paragraph was mirrored and the adjacent new one was not.

While mirroring I corrected two things in both copies:

- The original wording blamed "the POSIX shell on Linux". It happens on **macOS too** — `/bin/sh` has no `globstar` on either, and that is exactly where this was caught: the unquoted glob collapsed `apps/electron-backend/**/*.ts` to a single file (`src/main.ts`) on darwin.
- Added the part that actually explains why this survived so long: **the target still exits 0**, so a broken glob silently hides coverage rather than failing. Plus the check that catches it — compare the linted file count against `find <project> -name '*.ts' | wc -l` (245 vs 245 on this branch now; it was 1 vs 245 before).

4gray:
https://github.com/greptileai
@codex review

PR #1178, opened 2026-07-12
fix(window-controls): derive window-state pushes from events so controls reappear after fullscreen

4gray:
## What

After exiting video (HTML-element) fullscreen on Windows, the custom minimize/maximize/close controls stayed hidden forever: `app-window-controls` kept its `is-hidden` class even though the window had left fullscreen.

## Why

`attachWindowStateEvents` in `apps/electron-backend/src/app/app.ts` read `win.isFullScreen()` at event time. On Windows the transition isn't finished while `leave-full-screen` fires for an HTML fullscreen exit, so the renderer received a stale `{ isFullScreen: true }` push on exit — and no later event ever corrected it. Reproduced live on a Windows 11 dist build; the same race also fired on enter (stale `false`, controls not hiding during fullscreen).

## How

- Each `WINDOW:STATE_CHANGED` push now derives the flag its event names from the event itself: `enter/leave-full-screen` → `isFullScreen: true/false`, `maximize/unmaximize` → `isMaximized: true/false` (the other flag is still read best-effort and gets corrected by its own events).
- Also wires `enter/leave-html-full-screen` for platforms/triggers where the plain variants don't fire for element fullscreen; duplicate pushes with the same payload are harmless.

## Regression coverage

- `apps/electron-backend/src/app/app-window-state.spec.ts`: 6 unit tests on the event wiring — 4 fail against the old polling handlers.
- `apps/electron-backend-e2e/src/window-controls.e2e.ts`: new E2E case toggling HTML element fullscreen via a real click and asserting the controls lose `is-hidden` after exit — fails on the old build, passes with the fix. Skipped on window-managerless Linux CI like the file's other cases; runs on Windows/macOS-excluded matrix as before.

## Validation

- Manual on Windows 11 dist build (`ELECTRON_IS_DEV=0` + CDP): imported a raw-M3U playlist pointing at a locally served stream, playback running, toggled the Video.js fullscreen button on/off — controls hide during fullscreen and return after exit (`className=""`, bridge `{ isFullScreen: false }`).
- `pnpm nx run electron-backend-e2e:e2e-ci--src/window-controls.e2e.ts`: 5/5 passed.
- `pnpm nx test electron-backend`: 506 passed; the 13 failures are the pre-existing Windows-host artifacts in untouched suites (same set fails on master on this machine).
- Lint: changed files clean.

Docs: updated the `WINDOW:STATE_CHANGED` contract section in `docs/architecture/workspace-shell.md` to document the event-derived rule and the html-fullscreen events.

### DIFF
--- a/.changes/window-controls-fullscreen-exit.md
+++ b/.changes/window-controls-fullscreen-exit.md
@@ -5,4 +5,5 @@ area: window-controls
 
 On Windows, the minimize, maximize and close buttons disappeared for good after
 leaving fullscreen video playback — the only way to get them back was to restart
-the app. They now reappear as soon as you exit fullscreen.
+the app. They now reappear as soon as you exit fullscreen, and the maximize
+button no longer gets stuck on the wrong icon afterwards.
--- a/.changes/window-controls-fullscreen-exit.md
+++ b/.changes/window-controls-fullscreen-exit.md
@@ -1,0 +1,8 @@
+---
+type: fix
+area: window-controls
+---
+
+On Windows, the minimize, maximize and close buttons disappeared for good after
+leaving fullscreen video playback — the only way to get them back was to restart
+the app. They now reappear as soon as you exit fullscreen.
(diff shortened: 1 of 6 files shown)


4gray:
@greptileai

PR #1186, opened 2026-07-14
fix(settings): restore compact save action chip

SalemOurabi:
## What changed

- restore the Settings save action as a compact floating chip
- keep the chip aligned to the trailing edge with its original rounded border treatment
- reserve enough scroll space so the chip does not cover the final settings row
- add an Electron E2E regression test for the compact width and edge alignment

## Root cause

The save action had been changed from a floating chip into a full-width sticky footer. That made its blur surface much larger than the action itself and removed the intended spacing from the bottom edge.

## User impact

The save action remains sticky and easy to reach, but its background and blur are limited to a compact chip instead of spanning the entire Settings content width.

## Validation

- `pnpm nx run electron-backend-e2e:e2e-ci--src/settings-layout.e2e.ts` (1 passed on the isolated fix branch)
- `pnpm nx run electron-backend-e2e:e2e-ci--src/settings.e2e.ts` (8 passed before branch extraction)
- `pnpm nx lint electron-backend-e2e` (0 errors; existing warnings only)
- `git diff --check`

### DIFF
--- a/apps/electron-backend-e2e/src/settings-layout.e2e.ts
+++ b/apps/electron-backend-e2e/src/settings-layout.e2e.ts
@@ -13,6 +13,9 @@ test.describe('Electron Settings Layout', () => {
         const app = await launchElectronApp(dataDir);
 
         try {
+            await app.electronApp.evaluate(({ BrowserWindow }) => {
+                BrowserWindow.getAllWindows()[0]?.setSize(1000, 760);
+            });
             await openSettings(app.mainWindow);
 
             const layout = app.mainWindow.locator('.settings-layout');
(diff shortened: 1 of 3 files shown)


SalemOurabi:
@greptileai Please review this pull request, with particular attention to the compact Settings save-action chip, sticky positioning, blur surface sizing, content overlap, and Electron layout regressions.

SalemOurabi:
@greptileai please re-review the latest settings follow-up. The responsive layout now preserves sticky action spacing at <=1100px, with deterministic Electron E2E coverage at 1000x760.

PR #1204, opened 2026-07-18
fix(m3u): parse playlists with URLs longer than 2084 characters

### LINKED ISSUE #1189: Duvida sobre m3u?
Could someone please tell me why Iptvnator can't handle the .m3u files from this project: https://github.com/OwnerPlugins/pluto-tv-m3u It might be an issue with my computer, but it's strange because Hypnotix works perfectly with any of the files from that project. However, Iptvnator's interface is unbeatable for IPTV on Linux. It simply doesn't work, and there's no error output in the terminal; whether using mpv, VLC, HTML, or whatever player, it just won't play. The .m3u file contains dozens of channels, yet Iptvnator treats it as if it were only a single channel. Another confusing thing is that if I take any address from the list and run it directly in the terminal using standalone mpv, it doesn't work either. Yet Hypnotix plays the channels and uses mpv—though perhaps it uses its own version of mpv. Anyway, I don't have another computer to test this on, so I'd appreciate it if someone else could give it a try. Thank you for your attention.

4gray:
## Summary

Fixes #1189 — Pluto TV style playlists collapsed into a single channel.

**Root cause:** every stream URL in those playlists embeds a session JWT (~2200 chars). `validator.isURL` inside `iptv-playlist-parser` rejected anything over its IE-era 2084-char default limit, and the parser's stalled item index then made each next `#EXTINF` overwrite the same slot — the whole playlist collapsed into one channel (the last one).

**Fix:** synced [4gray/iptv-playlist-parser](https://github.com/4gray/iptv-playlist-parser) with upstream v0.15.2, which removes URL validation entirely and adds an explicit branch so `#` comments / unknown directives are never treated as URLs. Two fork deltas preserved on top (now documented in the fork README):

- **`radio` attribute** — load-bearing for the radio player, EPG suppression, external-player gating (15+ consumers)
- **pipe stripping** — `item.url` is cut at the first `|` (iptvnator feeds `item.url` verbatim to hls.js/mpv/vlc/catch-up/url-keyed favorites); `|User-Agent=`/`|Referer=` still land in `item.http`

Also dropped the now-dead `validator`/`is-valid-path` deps from the fork (zero runtime deps now) and pinned the dependency by commit SHA instead of tracking the fork's default branch.

**Second commit — optimized `parse()` (2.5–3× faster):** profiling showed the real hotspots were `getParameter`'s catastrophic `/^(.*)|/` backtracking (~39%) and 14+ `new RegExp` constructions per `#EXTINF` line (~25%), not the validator (~5%). The fork's `parse()` was rewritten around one precompiled attribute scan per line: 100k channels ~780 ms → ~285 ms, 10k ~79 ms → ~25 ms. Matters most for the PWA, where TEXT/FILE imports parse on the renderer main thread.

## Validation

- Fork test suite: 12/12 (3 new regression tests: >2084-char URLs, comments-not-urls, radio value)
- Optimized parser differential-verified **byte-identical** vs reference on 31 inputs: all test fixtures, 10 real Pluto playlists (live+VOD, 6 countries, 3123 channels), 10k/50k/100k synthetic corpora, adversarial samples
- New contract spec `apps/web/src/app/iptv-playlist-parser.contract.spec.ts`: 5/5 (fails on the old parser)
- `nx test m3u-utils` 48/48, `electron-backend` playlist specs 15/15, `web-backend` 15/15, `services` 34/34 — rerun green against the final pin
- E2E `web-e2e:e2e-ci--src/basic.e2e.ts` (M3U import flow): 6/6 — rerun green against the final pin
- `nx lint web`, `nx lint services` clean

### DIFF
--- a/apps/web/src/app/iptv-playlist-parser.contract.spec.ts
+++ b/apps/web/src/app/iptv-playlist-parser.contract.spec.ts
@@ -1,0 +1,100 @@
+import { parse } from 'iptv-playlist-parser';
+
+/**
+ * Contract tests for the 4gray/iptv-playlist-parser fork (jest maps the
+ * module to the real parser source via test-stubs/iptv-playlist-parser.mjs).
+ * Guards the fork-specific behaviors iptvnator depends on:
+ * - no URL length/format validation (issue #1189: Pluto JWT URLs > 2084 chars)
+ * - '#' comments never become URLs and never shift the item index
+ * - the fork-only `radio` attribute survives upstream syncs
+ * - `url` is stripped at the first '|' while pipe params land in `http.*`
+ */
+describe('iptv-playlist-parser contract (4gray fork)', () => {
+    const plutoUrl = (id: string) =>
+        `https://cfd-v4-service-channel-stitcher-use1-1.prd.pluto.tv/v2/stitch/hls/channel/${id}/master.m3u8?jwt=${'e'.repeat(2100)}&masterJWTPassthrough=true`;
+
+    it('parses playlists whose URLs are longer than 2084 characters (#1189)', () => {
+        const playlist = [
+            '#EXTM3U',
+            '#EXTINF:-1 group-title="Anime & Geek" tvg-id="one" tvg-name="Beyblade",Beyblade',
+            plutoUrl('one'),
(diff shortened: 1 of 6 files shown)


4gray:
@codex @greptileai

PR #1211, opened 2026-07-19
feat(matching): strip appended language/quality tags in title normalization

4gray:
## What

Follow-up to #1162 (PR-1 of the tag-extraction track). Teaches `normalizeTitleKeys()` — the single choke point feeding TMDB matching, `DB_MATCH_TITLES` trigram verification, similar rails, trending, and actor pages — to strip the appended language/quality tags that real portals put on VOD/series names.

## Why

Global search for one show returns 50–126 tagged variants (`|ALB| Fallout`, `4K-DE - The Pitt (2025) (US)`, `Breaking Bad-eng`, `Fallout_esp`). Measured on 248 real catalog names from four shows: **~35% normalized to polluted keys**, silently skipping TMDB enrichment (and getting cached as negative matches) and staying invisible to cross-portal matching.

## New handled forms (all conservative, casing/vocabulary-gated)

| Form | Example | Guard |
|---|---|---|
| Wrapped pipe tags | `\|ALB\| Fallout` | tag must contain a letter |
| Longer/compound leads | `EXYU\| X`, `4K-DE - X`, `4K-OSN+ - X` | dash/pipe only; colon stays 2–3 chars (`NCIS: LA` untouched) |
| Underscore suffixes | `X_eng`, `(US)_msub` | single underscore only (`The_Last_of_Us` intact) |
| Double-dash suffixes | `X--esp` | always safe |
| Joined dash tags | `X-DE`, `X-eng` | vocabulary + case-uniform (`Spider-Man`, `Kick-It` untouched) |
| Bare trailing tags | `X (2025) DE`, `Breaking Bad ES` | UPPERCASE vocabulary, skipped for ALL-CAPS titles (`Rocky II`, `Made in USA`, `Making It` untouched) |

Numeric fragments are never tags (`1917 - Behind the Lines` intact). The display-side `stripCountryPrefix()` learns the same compound/`+` prefixes and numeric guard. `buildSearchLookupKey()` gets a `|v2` suffix to invalidate cached negative match resolutions keyed on old polluted titles.

## Results (248 real names, four corpora committed as spec fixtures)

| Corpus | Clean matching keys before | after |
|---|---|---|
| The Pitt (55) | 67% | 96% |
| Fallout (97) | 62% | 99% |
| The Last of Us (52) | 68% | 100% |
| Breaking Bad (44) | 69% | 100% |
| **Overall** | **~65%** | **99%** |

Remaining misses are by design: lowercase unanchored tags (`X (2025) esp`) and foreign-script tails — stripping those would risk real titles. Localized subtitles (`Breaking Bad: A Química do Mal`) are intentionally untouched (indistinguishable from legitimate subtitles).

## Validation

- `shared-interfaces` 55/55, `m3u-utils` 51/51 (new corpus fixtures + guard cases)
- Downstream consumers all green: `services` (64), `portal-xtream-feature` (156), `portal-xtream-data-access` (165), `portal-stalker-data-access` (43), `workspace-dashboard-data-access` (144), `electron-backend` (756)
- Lint clean on all touched projects; no FTS reindex/migration needed (index stores raw titles)

Next up (separate PR): variant grouping in global search keyed on these normalized titles.

### DIFF
--- a/libs/services/src/lib/tmdb/tmdb-matcher.spec.ts
+++ b/libs/services/src/lib/tmdb/tmdb-matcher.spec.ts
@@ -78,10 +78,10 @@ describe('extractYear', () => {
 describe('lookup keys', () => {
     it('builds stable search and details keys', () => {
         expect(buildSearchLookupKey('the matrix', 1999)).toBe(
-            'title:the matrix|year:1999'
+            'title:the matrix|year:1999|v2'
         );
         expect(buildSearchLookupKey('the matrix', null)).toBe(
-            'title:the matrix|year:'
+            'title:the matrix|year:|v2'
         );
         expect(buildDetailsLookupKey(603)).toBe('id:603|v2');
     });
--- a/libs/services/src/lib/tmdb/tmdb-matcher.ts
+++ b/libs/services/src/lib/tmdb/tmdb-matcher.ts
@@ -70,7 +70,10 @@ export function buildSearchLookupKey(
     normalizedTitle: string,
     year: number | null
 ): string {
-    return `title:${normalizedTitle}|year:${year ?? ''}`;
+    // v2: normalizeTitleKeys learned to strip appended language/quality
+    // tags; the version suffix invalidates cached (incl. negative) match
+    // resolutions keyed on the old polluted titles
+    return `title:${normalizedTitle}|year:${year ?? ''}|v2`;
 }
 
 export function buildDetailsLookupKey(tmdbId: number): string {
--- a/libs/shared/interfaces/src/lib/title-normalization.util.spec.ts
+++ b/libs/shared/interfaces/src/lib/title-normalization.util.spec.ts
@@ -83,6 +83,11 @@ describe('provider tag stripping', () => {
         );
     });
 
+    it('keeps bare 4-5 char words before a spaced dash (real titles)', () => {
+        expect(normalizeTitle('DUNE - Part Two')).toBe('dune part two');
+        expect(normalizeTitle('ALIEN - Covenant')).toBe('alien covenant');
+    });
+
     it('strips underscore and double-dash suffix tags', () => {
         expect(normalizeTitle('Fallout_eng')).toBe('fallout');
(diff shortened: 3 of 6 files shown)


4gray:
Thanks for the review — addressed the three correctness findings in 0aebb88:

- **Underscore suffix not vocabulary-gated (P1):** now gated on the tag vocabulary, so `Mr_Robot` → `mr robot`, `Cowboy_Bebop` → `cowboy bebop`, `Mrs_Davis` → `mrs davis` are preserved. All corpus suffixes (`_eng`, `_esp`, `_sub`, `_tr`, `_it`, `_msub`) are in the vocabulary, so no coverage lost.
- **Expanded prefix accepts any uppercase word (P2):** single-segment tags before a *spaced dash* are back to 2–3 chars; only hyphen-compounds (`4K-DE`, `AR-SUBS`, `4K-OSN+`) and pipe-tags (`EXYU|`, `MULTI|`) may be wider. `DUNE - Part Two` and `ALIEN - Covenant` stay intact. The corpus only ever has `EXYU`/`MULTI` with pipes and all spaced-dash tags ≤3 chars, so coverage is unchanged. The display-side mirror got the same narrowing.
- **Hyphenated word endings (P2):** `IN` is excluded from the weak joined-dash/underscore paths, so `drive-in`/`Plug-in` are preserved; India still strips via the strong `IN| ` / `IN - ` forms.

Counter-example tests added for all of these; corpus coverage held at 99% (245/248).

On the **orphaned v1 cache rows (P2)** — valid, but a one-off `tmdb_metadata` purge is a DB-worker/appState migration that deserves its own focused change rather than riding along in a pure-normalization PR; the `|v2` key already guarantees correctness (old rows are simply never read again). I'll track it as a follow-up. Happy to fold it in here instead if you'd prefer.

4gray:
@greptileai please re-review — the last two commits (ES2015-safe trim fix and the tag-stripping false-positive guards for Mr_Robot / DUNE - Part Two / drive-in) address the earlier findings.

4gray:
@chatgpt-codex-connector

PR #1214, opened 2026-07-19
fix(epg): harden three latent edges from the #1165 manual-mapping review

### LINKED ISSUE #1: Cannot Play mpegts stream
Hi, application is works just fine with hls streams but when the play list is mpegts, video player just shows loading icon and never plays the video. On the other hand, stream seems connected and when i check it from the stream panel, i can see that i am connected to selected stream. Do you have any idea about how to play mpegts stream in the player?

4gray:
Follow-up to #1165 (manual EPG-to-channel mapping). During review of #1173 (Stalker mapping), the bots (Greptile + Codex) flagged four issues — all in **already-merged #1165 code**, not the Stalker delta. I verified each against `master`; three are real (minor) latent bugs and are fixed here. The fourth (Greptile "source collapse" symptom) turned out non-reproducible in the wired query paths, but shares its root cause with fix #1, which resolves it anyway.

### 1. EPG program dedup ignored `source_url`
The unique index and upsert conflict key `(channel_id, start, title)` collapsed programmes imported from **different XMLTV sources** that happened to share those columns, and the upsert reassigned `source_url` to the last importer. Since the query layer scopes by `source_url` and source-scoped deletes key on it, a programme could vanish from a source's guide or be deleted when another source was cleared.
**Fix:** key and index now include `source_url`. Migrated via a new `idx_epg_programs_dedup_v2` that drops the old source-blind index; the upsert no longer overwrites `source_url`. The changed-source-URL re-import case (the original reason source_url was omitted) at worst leaves harmless orphan rows under the old URL, which are no longer queried.

### 2. Xtream `getMapping` fallback truncated to 5 unordered streams
The provider-`epg_channel_id` fallback fetched only the first five streams sharing that id (`LIMIT 5`, unordered), so a mapping saved under a later matching stream was silently ignored.
**Fix:** replaced the two-step fetch-then-`inArray` with a single `content ⋈ categories ⋈ epg_channel_mappings` join that finds a mapping under **any** matching stream — no arbitrary cap, no `SQLITE_LIMIT_VARIABLE_NUMBER` exposure.

### 3. Xtream mapping dialog didn't refresh after close
Unlike the Stalker path (added in #1173), the Xtream dialog opened without an `afterClosed` refresh, so a remapped visible/selected channel kept its stale preview until the 5-minute TTL or a rescroll.
**Fix:** added `EpgQueueService.invalidate(streamId)` and a before/after mapping comparison in the channel-list dialog flow; when the mapping actually changed, the cached preview/resolution is dropped and the current viewport refetched.

### Tests
- `epg-database.spec`: source-aware dedup index/upsert + old-index drop assertions
- `epg-query.service.spec`: join-based `getMapping` resolves a mapping regardless of stream position, `LIMIT 1`, two inner joins
- `epg-queue.service.spec`: `invalidate()` clears cache/failure/ephemeral maps so the stream refetches

Build (web + electron-backend), affected unit suites (electron-backend, portal-xtream-data-access 157, portal-xtream-feature 144) and lint all green.

### DIFF
--- a/apps/electron-backend/src/app/events/epg-query.service.spec.ts
+++ b/apps/electron-backend/src/app/events/epg-query.service.spec.ts
@@ -417,7 +417,7 @@ describe('EpgQueryService', () => {
         });
         joinChain.where = jest.fn((condition: unknown) => {
             whereCalls.push(condition);
-            return { limit: joinLimit };
+            return { orderBy: jest.fn(() => ({ limit: joinLimit })) };
         });
 
         // Program lookup for the resolved (mapped) channel id.
--- a/apps/electron-backend/src/app/events/epg-query.service.spec.ts
+++ b/apps/electron-backend/src/app/events/epg-query.service.spec.ts
@@ -66,13 +66,32 @@ function createLimitedSelectChain(
     limitResult: unknown,
     whereCalls: unknown[]
 ): { from: jest.Mock } {
-    const limit = jest.fn().mockResolvedValue(limitResult);
+    const chain: Record<string, jest.Mock> = {
+        limit: jest.fn().mockResolvedValue(limitResult),
+    };
+    // groupBy/orderBy are optional links in the various query shapes; each
+    // returns the chain so where().groupBy().orderBy().limit() (and any
+    // subset) resolves to the same data.
+    chain.groupBy = jest.fn(() => chain);
+    chain.orderBy = jest.fn(() => chain);
     const where = jest.fn((condition: unknown) => {
         whereCalls.push(condition);
-        return { limit };
+        return chain;
     });
     return {
         from: jest.fn(() => ({ where })),
+    };
+}
+
+/** Program-query chain: from().where().groupBy().orderBy().limit(). */
+function createProgramChain(rows: unknown): { from: jest.Mock } {
+    const chain: Record<string, jest.Mock> = {
+        limit: jest.fn().mockResolvedValue(rows),
+    };
+    chain.groupBy = jest.fn(() => chain);
+    chain.orderBy = jest.fn(() => chain);
+    return {
+        from: jest.fn(() => ({ where: jest.fn(() => chain) })),
     };
 }
 
@@ -434,15 +453,6 @@ describe('EpgQueryService', () => {
             episodeNum: null,
             sourceUrl: null,
         };
-        const programChain = {
-            from: jest.fn(() => ({
-                where: jest.fn(() => ({
-                    orderBy: jest.fn(() => ({
-                        limit: jest.fn().mockResolvedValue([programRow]),
-                    })),
-                })),
-            })),
-        };
 
         const select = jest
             .fn()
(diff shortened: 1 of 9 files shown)


4gray:
Addressed all four review comments in `47e4171f`:

- **playlistId re-read after await** (Greptile ×2): the dialog flow now takes the already-validated `playlistId` as a parameter instead of re-reading `currentPlaylist()` after the `getEpgMapping` roundtrip, so navigation during the IPC can't open the dialog with `playlistId: undefined`.
- **in-flight request restoring an invalidated preview** (Codex): `invalidate()` now bumps a per-stream epoch and clears `inFlight`; `fetchEpg` checks the epoch before writing, so a request already running when the mapping changes has its stale result discarded, and the immediate re-enqueue can schedule a fresh mapping-aware fetch. Added a regression test.
- **playlist scope in the shared-epg_channel_id join** (Codex): this backend layer only receives the provider `epg_channel_id`, not the caller's playlist, so it can't scope to the caller's own mapping — the renderer already resolves the playlist-scoped key directly and this is a best-effort fallback. Added deterministic `ORDER BY (playlist_id, xtream_id)` before `LIMIT 1` so the choice is stable rather than storage-order dependent, and documented the limitation.

Affected suites green (electron-backend 775, portal-xtream-data-access 158, portal-xtream-feature 144), lint clean, web build clean.

@greptileai please re-review.

4gray:
Pushed `e9d90120` on top of the review fixes — split the oversized invalidation spec so lint is green again. Current head has all four review comments addressed plus the lint fix.

@greptileai please re-review the latest commit.
@chatgpt-codex-connector please re-review the latest commit.

4gray:
Both second-order P2s addressed in `25210927`:

- **Unscoped timeline dedup**: added `toEpgPrograms()` which collapses duplicate `channel|start|title` slots after mapping/validation, applied at every `getChannelPrograms` return — so the M3U timeline (queried without `sourceUrls`) no longer shows a programme twice even though the source-aware index now keeps per-source rows.
- **In-flight marker ownership**: `fetchEpg`'s `finally` now releases the marker only when the completing request still owns it (epoch unchanged), so a request that went stale after `invalidate()` can't drop a re-enqueued request's marker and let a third concurrent fetch start.

Regression tests added for both. Full `pnpm run lint` (exit 0), web build, electron-backend (776) and portal-xtream-data-access (159) suites all green.

@greptileai @chatgpt-codex-connector please re-review.

4gray:
Moved the dedup into SQL in `7a86485b` so it runs before the row limits:

- **Timeline** (`selectChannelPrograms` / legacy variant): `GROUP BY (channel_id, start, title)` before `ORDER BY start LIMIT 500`, so the cap counts distinct programmes, not duplicate cross-source rows.
- **Batch current programmes** (`selectCurrentProgramsForChannelIds`): `GROUP BY channel_id` before `LIMIT channelIds.length`, so duplicate current slots for one channel can't starve another channel's preview.

The JS `toEpgPrograms()` dedup stays as a safety net (e.g. legacy NULL-source rows, which the unique index treats as distinct). Query-chain test mocks updated for the new groupBy link; electron-backend suite (776) green, full lint (exit 0) and electron-backend build clean.

@greptileai @chatgpt-codex-connector please re-review.

PR #1216, opened 2026-07-20
fix(catchup): parse negative sub-hour XMLTV offsets correctly

SalemOurabi:
## Summary

XMLTV timestamps carrying negative sub-hour timezone offsets (e.g. `202604100800 -0030`) were resolved as if they were UTC. The hour component of such an offset is the string `-00`, which evaluates to `-0`, and `Math.sign(-0)` returns `-0` — so the minutes' sign was dropped and the whole offset collapsed to zero. Catch-up URLs built from these programs pointed 30 minutes (or whatever the sub-hour offset was) away from the real program start.

The fix derives the sign directly from the offset string and applies it to the combined hour+minute magnitude, so `-0030` now correctly shifts by −30 minutes while `+0530`, `-0000`, and full-hour offsets keep their previous behavior.

## Changes

- `libs/shared/m3u-utils/src/lib/catchup.utils.ts`: sign-safe XMLTV offset conversion in `getEpgProgramTimestampSeconds`
- `libs/shared/m3u-utils/src/lib/catchup.utils.spec.ts`: regression tests for `-0030` (fails on the old implementation), `+0530`, and `-0000`

## Validation

- `node tools/testing/run-web-esm-lib-tests.mjs libs/shared/m3u-utils` — 3 suites, 55 tests passing
- `pnpm nx lint m3u-utils` — clean

### DIFF
--- a/libs/shared/m3u-utils/src/lib/catchup.utils.spec.ts
+++ b/libs/shared/m3u-utils/src/lib/catchup.utils.spec.ts
@@ -251,6 +251,24 @@ describe('catchup.utils', () => {
         );
     });
 
+    it('applies positive sub-hour XMLTV offsets such as +0030', () => {
+        // Regression: Number('+00') === 0, so a Math.sign()-based conversion
+        // dropped the minutes and treated +0030 as UTC.
+        expect(
+            resolveM3uCatchupUrl(
+                baseChannel,
+                {
+                    ...archivedProgram,
+                    start: '202604100800 +0030',
+                    startTimestamp: null,
+                },
+                1_775_820_000
+            )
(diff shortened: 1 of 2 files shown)


SalemOurabi:
@greptileai please review

4gray:
@greptileai please re-review — commit 2f2cd586 adds the requested +0030 regression coverage.

4gray:
@codex review

4gray:
Thanks @SalemOurabi for the focused and useful fix! We added coverage for the symmetric `+0030` case, and all checks and reviews are now green. Merging this in 🙌

PR #1232, opened 2026-07-25
fix(m3u): avoid persisting hydrated favorites

4gray:
## Summary

Prevent route-driven M3U favorites hydration from triggering a second full playlist persistence cycle.

## Root cause

`M3uWorkspaceRouteSession` loaded the playlist and its favorites from SQLite, then dispatched `FavoritesActions.setFavorites`. That action is also the user-mutation action observed by the persistence effect, so simply opening an imported playlist caused the full M3U payload to be read and written again.

## Changes

- add reducer-only `FavoritesActions.hydrateFavorites`
- use hydration for both successful route loads and the load-error cleanup path
- keep `setFavorites` for real user-driven reorder persistence
- keep `updateFavorites` for favorite toggles
- add reducer and route-session regression coverage, including the error path
- document the hydration/persistence boundary

## Performance evidence

The same local synthetic M3U benchmark was run before and after the production change, with one warm-up and five measured runs per size:

| Channels | Before median | After median | Improvement |
| ---: | ---: | ---: | ---: |
| 10,000 | 1,040.2 ms | 762.3 ms | 26.72% |
| 50,000 | 4,287.5 ms | 3,127.8 ms | 27.05% |
| 100,000 | 8,912.1 ms | 6,400.4 ms | 28.18% |

The measured route-load database work changed from two playlist reads plus two full upserts to one read plus one upsert. Benchmarks used only generated local playlists; no provider URLs or credentials were used. Raw profiles and traces remain untracked under `dist/performance/`.

## Validation

- `pnpm nx test m3u-state --runInBand --skipNxCache` — 35 passed
- `pnpm nx test playlist-m3u-feature-player --runInBand --skipNxCache` — 43 passed
- `pnpm nx lint m3u-state --skipNxCache` — passed (3 pre-existing warnings)
- `pnpm nx lint playlist-m3u-feature-player --skipNxCache` — passed
- `pnpm nx run electron-backend-e2e:e2e-ci--src/favorites.e2e.ts --skipNxCache` — 6 passed

The new error-path regression was also verified red/green: it fails when the catch path is temporarily restored to `setFavorites` and passes with `hydrateFavorites`.

## Trade-off

Hydration is intentionally reducer-only. Callers that represent a user mutation must continue to dispatch `setFavorites` or `updateFavorites`; using `hydrateFavorites` for a mutation would update NgRx state without persisting it.

Legacy non-string favorite values are still filtered on read, but are no longer implicitly cleaned up by rewriting the entire playlist on every route entry. Any durable cleanup should be an explicit migration.

### DIFF
--- a/.changes/m3u-favorites-hydration.md
+++ b/.changes/m3u-favorites-hydration.md
@@ -1,0 +1,7 @@
+---
+type: perf
+area: m3u
+---
+
+Importing large M3U playlists is faster because IPTVnator no longer rewrites
+the entire playlist when loading saved favorites.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -430,7 +430,7 @@ State management via NgRx (`libs/m3u-state/`):
 - `PlaylistActions`: loadPlaylists, addPlaylist, removePlaylist, parsePlaylist
 - `ChannelActions`: setChannels, setActiveChannel, setAdjacentChannelAsActive
 - `EpgActions`: setActiveEpgProgram, setCurrentEpgProgram, setEpgAvailableFlag
-- `FavoritesActions`: updateFavorites, setFavorites
+- `FavoritesActions`: updateFavorites, setFavorites, hydrateFavorites
 
 See `docs/architecture/m3u-playlist-module.md` for complete documentation.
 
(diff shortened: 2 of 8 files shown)


4gray:
@greptileai

4gray:
@greptileai

4gray:
@codex review

PR #1239, opened 2026-07-25
fix(tmdb): stop a broken provider tmdb_id from suppressing enrichment

4gray:
First fix out of the [TMDB capability roadmap](https://github.com/4gray/iptvnator/pull/1238) (item A1) — a confirmed defect in shipped code, not a feature.

## The bug

`enrich()` trusted the provider's `tmdb_id` unconditionally:

```ts
const tmdbId =
    parseProviderTmdbId(query.tmdbId) ??
    (await this.resolveIdBySearch(mediaType, query));
```

IPTV panels ship broken ids, and this had two failure modes:

1. **Dead id** — a garbage-but-integer value short-circuits the title search entirely. `getDetails` 404s, the outer `catch` swallows it, and the item is left **permanently unenriched** (no plot, no cast, no artwork) for a title the search would have matched. Failed detail fetches cache nothing, so the wasted request repeats on every single re-open.
2. **Stale-but-valid id** — never throws, and nothing sanity-checks the resolved title, so we confidently render **another film's** plot, cast and poster.

Affects Xtream movies (`xtream-tmdb-enrichment.ts` is the only path passing a provider `tmdbId`) — the highest-traffic detail view.

## The fix

The provider id becomes a *hint*, not a verdict:

- id resolves and the title looks like ours → used as before, **one details call, no search** (unchanged happy path)
- id fails to resolve → title search gets its turn, and the id is negative-cached (7 days, one language-independent row) so the 404 is not repeated forever
- id resolves but the title matches none of the variants we would have searched → the search competes; if it produces a confident match, that wins and the id is remembered as bad

### The deliberate non-rejection

A title mismatch is treated as **suspicion, not proof**. TMDB returns titles in the *request* language, so a Russian provider title legitimately fails the name check against an `en-US` payload. If the search finds nothing confident, the provider payload is kept.

Net effect: this change can only **add** enrichment, never remove it.

## Refactor

Search resolution + the bad-id cache move into a new `TmdbIdResolverService`. `tmdb-enrichment.service.ts` was at 290 lines against the 300-line target (269 now), and the resolver is independently testable.

## Validation

- New `tmdb-enrichment.service.spec.ts` — 6 regressions: happy path issues exactly one details call and **no** search; 404 falls back to search; stale id is overridden by a confident match; **payload is kept when the title differs but the search finds nothing**; known-bad id skips the dead call entirely; no-id/no-match returns null
- `tmdb-matcher.spec.ts` — `detailsMatchProviderTitle` (noisy titles, original-title-only match, series `name`/`original_name`, genuine mismatch, empty inputs) and the namespaced bad-id cache key
- `nx test services portal-xtream-data-access portal-stalker-data-access` — 450 tests green
- `nx lint services` 0 errors, `nx build web` ✅

No i18n keys, no UI change, no cache-key bump.


4gray:

4gray:
@codex review

4gray:
@greptile review

4gray:
@codex review

4gray:
@greptile review

4gray:
@codex review

4gray:
@greptile review

PR #1241, opened 2026-07-25
perf(stalker): skip wasted series-seasons request for non-series items

4gray:
## Problem

`setSelectedItem` ([with-stalker-selection.feature.ts](https://github.com/4gray/iptvnator/blob/master/libs/portal/stalker/data-access/src/lib/stores/features/with-stalker-selection.feature.ts)) mirrored **every** selection's id into `selectedSerialId`, and `serialSeasonsResource` ([with-stalker-series.feature.ts](https://github.com/4gray/iptvnator/blob/master/libs/portal/stalker/data-access/src/lib/stores/features/with-stalker-series.feature.ts)) only gates on `selectedSerialId` + `currentPlaylist`.

So opening *any* Stalker detail page fired a pointless portal request:

```
get_ordered_list&type=series&movie_id=<id>
```

This happened for:

- plain VOD movies (no series data at all),
- vclub items with embedded `series[]` — the response is fetched and then **discarded** by `mapRegularSeriesSeasons`, which prefers the embedded episodes,
- Ministra `is_series` VOD items, which resolve seasons through the separate `type=vod` path,
- ITV channel selections in the live layout.

…on every entry path: browse, favorites, recent, dashboard and search.

## Fix

Set `selectedSerialId` only when `selectedContentType === 'series'`; every other content type clears it. That also prevents a stale serial id from lingering when a VOD item is opened right after a series.

**The gate is deliberately on content type alone, not item shape.** Under the `series` content type `serialSeasonsResource` is the *only* episode source — the detail templates render `<app-stalker-series-view />` with no `vodWithSeries` input ([stalker-catalog-detail.component.html:26-28](https://github.com/4gray/iptvnator/blob/master/libs/portal/stalker/feature/src/lib/stalker-catalog-detail/stalker-catalog-detail.component.html#L26-L28)), and `isVodSeries()` requires content type `vod`. Additionally gating on `is_series`/`series[]` would leave a series-section item that happens to carry either field with a silently empty episode list (no spinner, no error). It would also buy nothing: every embedded-series and Ministra item is opened under content type `vod` (`resolveDetailMode`, the search filter, the route section), so the content-type check alone already suppresses 100% of the wasted requests.

The first commit had that unnecessary conjunct; the third commit removes it after self-review. The resource loader already early-returns `[]` on an empty id without issuing a request, so it needed no change.

## Tests

- `with-stalker-selection.feature.spec.ts` — serial id set for a regular series; not set for plain VOD, embedded-`series[]` and Ministra `is_series` items (each paired with the VOD content type they actually occur under); cleared when a non-series item follows a series; **and still set** for a `series` selection carrying `is_series` or `series[]`.
- `with-stalker-series.feature.spec.ts` (new) — composes the real selection + series features with a mocked `DataService` and asserts at the **request level**: exactly one `type=series` fetch for a regular series, zero portal calls for plain VOD and embedded-series selections, no `type=series` call for a Ministra item (its legitimate `type=vod` season request is still allowed), no refetch when a VOD selection follows a series, and a regression guard that a `series` selection carrying `is_series` still fetches.

Regression check: with the fix reverted to the `origin/master` implementation, the new tests fail (8 failures); with it applied, all 107 pass.

## Validation

| Command | Result |
| --- | --- |
| `pnpm nx test portal-stalker-data-access` | 17 suites, 107 tests passed |
| `pnpm nx test portal-stalker-feature` | 9 suites, 77 tests passed |
| `pnpm nx lint portal-stalker-data-access` | clean |

## Docs

`docs/architecture/stalker-store-api-baseline.md` claimed all three selection ids are synchronized in `setSelectedItem`, which is no longer true. Both that file and the VOD/Series modes section of `docs/architecture/stalker-portal.md` now record the actual invariant and why the gate must stay on content type.

### DIFF
--- a/docs/architecture/stalker-portal.md
+++ b/docs/architecture/stalker-portal.md
@@ -271,10 +271,14 @@ Stalker has multiple real-world data shapes. The current implementation supports
 
 - Seasons come from API resource (`serialSeasonsResource`).
 - Episodes are derived from season payload.
-- This is the only shape that sets `selectedSerialId`, which is what drives
-  `serialSeasonsResource`. Modes 2 and 3 below carry their episodes elsewhere,
-  so selecting them must leave the id unset — otherwise every detail open
-  fires a `get_ordered_list&type=series` request whose result is discarded.
+- This is the only mode that sets `selectedSerialId`, which is what drives
+  `serialSeasonsResource`. It is set purely from `selectedContentType ===
+'series'` — the `series` detail branch renders `<app-stalker-series-view />`
+  with no `vodWithSeries` input, so the API resource is its only episode
+  source and the fetch must never be gated on item shape.
+- Modes 2 and 3 below are always opened under the `vod` content type, which
+  leaves the id unset — otherwise every VOD detail open would fire a
+  `get_ordered_list&type=series` request whose result is discarded.
 
 2. VOD with Embedded `series[]`:
 
--- a/docs/architecture/stalker-portal.md
+++ b/docs/architecture/stalker-portal.md
@@ -271,6 +271,10 @@ Stalker has multiple real-world data shapes. The current implementation supports
 
 - Seasons come from API resource (`serialSeasonsResource`).
 - Episodes are derived from season payload.
+- This is the only shape that sets `selectedSerialId`, which is what drives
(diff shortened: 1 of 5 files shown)


4gray:
@greptileai

PR #1242, opened 2026-07-25
fix(tmdb): series cast was the latest season only, not the show

4gray:
Roadmap item A2 — a defect in shipped code wearing a feature costume.

## The bug

TMDB documents a TV id's `credits` as **the credits of the latest season**. We requested exactly that and rendered it as "the cast", so on every long-running show we silently dropped every regular who had left. On a show like The Boys the chips list whoever happens to be in the newest season — not the ensemble the user is looking for.

## The fix

The TV details request now also appends `aggregate_credits`, which spans the whole run. But that one has the mirror-image gap — per TMDB it **omits the newest season** — so neither payload alone is the cast:

- `credits` alone → loses everyone who left
- `aggregate_credits` alone → loses everyone who just joined

`unifiedTvCast` unions them: whole-run billing order first, then newest-season-only arrivals appended, deduplicated by person id. Arrivals go last deliberately — their `order` is scoped to that season and would otherwise outrank show-level billing. Characters come from the aggregate's `roles[]` shape rather than a flat `character`.

Movies are untouched: `/movie/{id}` has no `aggregate_credits`, and its `credits` is already the full cast.

## No cache-key bump — on purpose

Rows cached before this simply lack `aggregate_credits` and fall back to the previous behaviour until they expire naturally. That's a deliberate call: the roadmap schedules **one** consolidated bump for when the remaining `append_to_response` additions (images, certifications, alternative_titles) land together, and bumping now would mean invalidating every user's details cache twice. The "Clear metadata cache" button (next PR) gives anyone who wants the fix immediately an escape hatch.

## Validation

- 5 new regressions: departed regulars retained, arrivals appended after show billing order, characters read from `roles[]`, no duplicates across the two payloads, graceful fallback for pre-aggregate rows
- `nx test services portal-xtream-data-access portal-stalker-data-access` — 443 tests green (existing merge tests pass unchanged, confirming movies and the fallback path are untouched)
- `nx lint services` 0 errors, `nx build web` ✅
- Docs updated


4gray:

4gray:
@codex review

4gray:
@greptile review

4gray:
@codex review

4gray:
@greptile review

4gray:
@codex review

4gray:
@greptile review

PR #1244, opened 2026-07-25
feat(tmdb): metadata cache panel with a clear button in settings

4gray:
Adds **Metadata cache — N entries · X MB** with a Clear button to `Settings > Metadata (TMDB)`, right next to the API key.

## Why there

The cache is TMDB's, not the app's in general — a user whose metadata misbehaves goes to the same section that already has "Check key". A separate storage section would have created a second place to look.

## What it's for

1. **Dropping wrong or stale metadata** so the next detail open refetches it — the practical answer to "this shows the wrong film" until the manual match override lands.
2. **Seeing what the cache costs** on disk.
3. **Reclaiming orphaned rows.** When a lookup-key version bump lands (the roadmap schedules one consolidated `append_to_response` bump), old rows become *unreachable*, not deleted. Nothing else would ever collect them.

## Implementation notes

- Sizing is a **full table scan** — `LENGTH()` on TEXT counts characters, so the `SUM` casts to BLOB to actually get bytes. That's why stats load **lazily, once the TMDB section becomes active**, rather than on every settings open.
- Clearing is always safe: enrichment refetches on demand, so the only cost is the next few requests. No confirmation dialog for that reason.
- **Both environments**: the PWA has no bridge, so the service reports and clears its session-scoped in-memory map instead of the SQLite table.
- Button disables itself at zero entries and while clearing.

## Validation

- `nx test web services electron-backend` green — includes both contract fixtures (preload bridge + DB-worker payload shapes) which enforce full coverage of new IPC methods
  - one unrelated flake (`download-runtime.spec.ts`, 34s timeout under parallel load) — passes in isolation, untouched by this change
- i18n: 4 keys × 19 locales via the `tools/i18n` workflow; coverage `missing: 0 / extra: 0` and placeholder-integrity check clean (`{{entries}}` / `{{size}}` preserved everywhere)
- `nx lint` 0 errors, `nx build web` ✅
- Docs updated


4gray:

4gray:
@codex review

4gray:
@greptile review

4gray:
@codex review

4gray:
@greptile review

4gray:
@codex review

4gray:
@greptile review

PR #1256, opened 2026-07-25
chore(release): author release notes in .changes instead of reconstructing them

4gray:
## Why

`CHANGELOG.md` has been frozen at **0.12.0 (2023-03-11)** while the app ships **0.23.0**. `semantic-release` + `@semantic-release/{changelog,git,npm}` + `conventional-changelog-cli` sat in `devDependencies` with **no config at all**. The GitHub release body is `generate_release_notes: true` — a flat list of PR titles. The notes users actually read are a hand-written 280-line MDX post, reconstructed from memory three months after the fact.

The missing piece was never version math. It was **authored, user-facing notes captured at PR time**.

## What

A `.changes/*.md` note per user-visible change, plus a generator that composes the three release surfaces from the accumulated notes.

```markdown
---
type: feature
area: playback
issues: [1187]
screenshot: up-next-rail
---

Series now show an "Up Next" rail beside the player on wide windows.
```

```bash
node tools/release/build-release-notes.mjs --validate
node tools/release/build-release-notes.mjs --version 0.24.0 --format github
node tools/release/build-release-notes.mjs --version 0.24.0 --format changelog
node tools/release/build-release-notes.mjs --version 0.24.0 --format blog
node tools/release/build-release-notes.mjs --version 0.24.0 --consume
```

Only `--consume` deletes anything; every other mode is a safe dry run.

## Why not Changesets

Changesets versions **multiple published packages**. This repo has exactly one private package with a version (`libs/**` have no `package.json` at all) and publishes nothing to npm. Its `changeset version` step would rewrite `CHANGELOG.md` into a flatter format than the blog post, and its semver bump fights a version that is chosen deliberately under electron-updater constraints. The valuable half — the note file — is what this PR takes.

`release-please` was the closest off-the-shelf fit but derives notes from commit titles, which is the exact problem being solved.

## Notable decisions

- **Hand-rolled frontmatter parser over a YAML engine.** The schema is closed, so the parser can reject unknown keys — that is what catches `scopr:` instead of `area:`. Zero new dependencies.
- **Authors never write a PR number.** They cannot know it while the branch is local; the generator resolves it from the commit that added the note.
- **MDX-significant characters are escaped.** A note body containing `<live>` would otherwise break the website build.
- **The blog scaffold ships `draft: true` with explicit `TODO` headings.** An earlier version derived headings from note bodies and produced a 140-character `###` on the first real example. Headings are editorial work; a visible TODO is more honest than pretending otherwise.
- **`CHANGELOG.md` is revived with a pointer**, not a backfill — 0.13.0–0.23.0 point at GitHub Releases and the website rather than inventing history.
- **Five dead packages removed** (−1662 lockfile lines).

## Follow-ups (not in this PR)

- **PR 2** — CI gate requiring a note when a PR touches `apps/**`/`libs/**` (with a `no-release-note` label escape), plus `release-notes` / `release-cut` agent skills.
- **PR 3** — version-agnostic screenshot capture driven by a manifest, with fail-closed mock-data gates. Full spec in `.plans/2026-07-25-release-notes-pipeline.md`.

## Test plan

| Check | Result |
| --- | --- |
| `node --test tools/release/release-notes.test.mjs` | 26/26 pass |
| `pnpm nx test release-tools` | pass |
| `pnpm nx lint release-tools` | pass |
| `pnpm nx build website` with a generated scaffold present | pass, 14 pages |
| All four CLI modes against real notes | `internal` excluded from the release body, collapsed in CHANGELOG |
| Invalid note | all four problems reported at once, exit 1 |
| Blog overwrite guard | refuses without `--force` |

Smoke-run artifacts (`CHANGELOG.md` insertion, `v0-24-release-notes.mdx`) were reverted/removed; the diff contains only intentional changes.

No release note for this PR itself: tooling with no user-visible effect, which its own rules classify as `no-release-note`.


4gray:
Review pass — two accepted, one clarified, one deferred with a design note.

**Greptile P1 — format scripts omit `--version`.** Real, fixed in b20f62b2, but not by adding the flag to each script. The release version already lives in the root `package.json`, and bumping it is the deliberate act that starts a release — so that is now the default source, with `--version` kept as an override for dry runs before the bump. The notice prints to **stderr**, so `--format github` still pipes cleanly.

**Greptile P1 — backslashes in MDX alt text.** Correct and accepted. A body ending in `\` produced an unterminated string literal and would have broken the website build. Backslashes are now escaped before apostrophes, with a regression test that I verified fails against the old code (deleted the escape line, watched it go red, restored).

**Codex P2 — patch releases share a slug.** Working as intended, now documented rather than changed. The website publishes **one post per minor version** (`v0-18` … `v0-22`), and screenshots live under the matching `blog/v0-24/` directory. A patch release edits that post instead of starting a second one — which is exactly what the "already exists" guard enforces. The failure mode read like a bug because the message did not say so; it now explains the convention, as do the slug helper and `.changes/README.md`.

**Codex P1 — generator not wired into the release workflow.** The observation is correct: `build-and-make.yaml` still leaves `release-meta.body` empty and relies on `generate_release_notes: true`, so authored notes do not reach a published release yet.

The implied fix does not work, though. By the time a `v*` tag exists, `--consume` has already deleted `.changes/` — the sequence is bump → generate → commit CHANGELOG + consume → tag → push, so the tag build has no notes left to render. The correct wiring reads what the generator already wrote: extract the `CHANGELOG.md` section matching the tag's version into `release-meta.body`, keep `generate_release_notes: true` so GitHub's commit list layers *below* the authored notes, and fail the release job when no matching section exists so a forgotten `--format changelog` cannot silently ship PR-title-only notes.

That is workflow surgery on the most delicate file in the repo and cannot be exercised until a real tag is pushed, so it stays out of this PR. It is now explicit PR 2 scope with the design written down in `.plans/2026-07-25-release-notes-pipeline.md`.

Tests: 27 passing.

4gray:
@greptileai review

4gray:
@codex review

PR #1257, opened 2026-07-25
ci(release): gate PRs on an authored release note

4gray:
## Why

#1256 landed the `.changes/` format and generator. Without enforcement the habit decays within a few PRs — this adds the gate that makes it stick, for maintainers, agents, and external contributors alike.

## What

**"Release note gate" CI job** (PR-only, in `ci.yml`):
1. Validates every `.changes/*.md` against the schema.
2. Requires an **added** note when the PR touches runtime code under `apps/` or `libs/` — or the `no-release-note` label (already created in the repo).

Auto-exempt: `*.spec.ts`, `*.e2e.ts`, `__snapshots__/`, `apps/website/`, `apps/*-e2e/`, `apps/*-mock-server/`, `libs/shared/testing/`, `*.md`. A PR that only changes those never needs a note or a label.

**Design choices:**
- Policy is a pure function in `tools/release/check-release-note-gate.mjs`, fed PR files+labels as JSON — 10 unit tests instead of untestable workflow bash. The failure message lists the exact triggering files and the exact fix.
- Labels are fetched **live**, not from the event payload — apply the label and re-run the check, no new push needed.
- The job is dependency-free Node: no `pnpm install`, runs in seconds.

**Skills** (committed to `.codex/skills/`, the repo's canonical skill location):
- `release-notes` — how to write a note, with good/bad examples and the exemption list.
- `release-cut` — the full release sequence: bump → review notes → changelog → blog scaffold → screenshots (mock servers only) → consume → tag → verify draft, plus failure modes (the tag build fails closed if the changelog step was skipped).

`CLAUDE.md`/`AGENTS.md` sections updated to reference the gate and skills.

## Test plan

| Check | Result |
| --- | --- |
| `pnpm nx test release-tools` | 47/47 (10 new gate cases: label escape, modified-note-doesn't-count, renamed-note counts, README exclusion, exemptions, mixed exempt+runtime, listing truncation) |
| Gate CLI smoke: fail / note / label paths | exit 1 / 0 / 0 with correct messages |
| Gate step shell | shellcheck at CI severity (`--severity=warning`), clean |
| `ci.yml` | YAML-parse verified; actionlint runs in CI itself |

This PR touches no runtime code (`ci.yml`, `tools/`, docs, skills), so by its own policy it needs no note — and the gate on this very PR should pass green as the first live demonstration. `no-release-note` label applied for explicitness.

Remaining from the plan: PR 3 — manifest-driven screenshot capture with fail-closed mock-data gates.

### DIFF
--- a/.claude/skills/release-cut/SKILL.md
+++ b/.claude/skills/release-cut/SKILL.md
@@ -1,0 +1,78 @@
+---
+name: release-cut
+description: Cut an IPTVnator release — bump the version, generate release notes from .changes/, scaffold the website post, tag, and verify the draft. Use when asked to release, cut a version, prepare release notes, or publish a new version.
+---
+
+# Release Cut
+
+The pipeline turns accumulated `.changes/*.md` notes into all three release
+surfaces. Order matters: **the tag build extracts the CHANGELOG section into
+the GitHub release body and fails if it is missing**, so the changelog step
+is not optional.
+
+## Sequence
+
+1. **Pick the version** — deliberate choice, edit `version` in the root
+   `package.json`. Bare semver only: any suffix flips electron-updater into
+   prerelease mode and leaks into installer version fields.
+
+2. **Review the notes** — read every file in `.changes/`. Fix wording (user
+   language, not reviewer language), then:
+
+    ```bash
+    pnpm run release:notes:validate
+    ```
+
+3. **Generate the changelog section** (idempotent per version — rerunning
+   replaces the section, so regenerate freely until it reads well):
+
+    ```bash
+    pnpm run release:notes:changelog
+    ```
+
+4. **Scaffold the website post**:
+
+    ```bash
+    pnpm run release:notes:blog
+    ```
+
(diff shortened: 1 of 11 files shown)


4gray:
Both findings accepted and fixed in d5afdd89.

**`renamed` no longer satisfies the gate.** Codex and Greptile independently flagged this and they are right. The PR files API compares base…head, so a note created *and* renamed inside one PR still reports as `added` — a `renamed` entry can only mean the file already existed on the base branch. Accepting it let a runtime-code PR pass by moving another PR's unconsumed note, which documents nothing and leaves the generator without an adding commit to resolve the PR link from.

**Nested notes no longer satisfy it either.** `loadNotes()` reads only the immediate `.changes/` directory, so `.changes/sub/note.md` passed the old prefix check while never being validated or rendered into any release surface — a silent hole exactly where the gate is supposed to be load-bearing. Now a direct child is required.

Both are covered by failing-gate tests (12 gate cases total).

4gray:
@greptileai review

PR #1258, opened 2026-07-25
fix(deps): patch transitive runtime CVEs via pnpm overrides

4gray:
Closes 13 runtime-scope Dependabot advisories that Dependabot **structurally cannot fix**: every vulnerable package here is transitive, so the bot has no lever until each parent publishes a release widening its own pin. `pnpm.overrides` is that lever.

## What changed

Five overrides, using the pinned-source form already established in this repo:

| Package | From → To | Advisories | Arrives via |
| --- | --- | --- | --- |
| `@xmldom/xmldom` | 0.8.11 → 0.8.13 | 5 high | `video.js` → `@videojs/http-streaming` → `mpd-parser` |
| `fast-uri` | 3.1.0 → 3.1.4 | 4 high | `electron-conf` → `ajv` |
| `js-yaml` | 4.1.1 → 4.3.0 | 1 high, 1 medium | `electron-updater` |
| `form-data` | 4.0.5 → 4.0.6 | 1 high | `axios` |
| `ajv` | 8.17.1 → 8.18.0 | 1 medium | `electron-conf` |

## The constraint that shaped the version choices

**Every target stays inside its parent's declared semver range.** For three of the five, the newest published version is *outside* that range — `xmldom` 0.9.x, `fast-uri` 4.x, `js-yaml` 5.x. Taking "latest" would have installed cleanly and broken at runtime. `docs/architecture/dependency-security-overrides.md` records this so the next person bumping these doesn't walk into it.

## Honest note on reachability

`@xmldom/xmldom` is the largest block by advisory count, but it is **not on the DASH playback path**. It arrives through Video.js's VHS bundle, while the app routes every `.mpd` to Shaka (`isDashChannel()` → `shaka-video-session.ts`), which uses its own DASH parser. The code ships in the bundle but nothing feeds it. This is defence in depth, not closing a live hole.

The genuinely reachable one is **`js-yaml`**: `electron-updater` uses it to parse `latest.yml` fetched from GitHub releases — remote input on a shipped path.

## Deliberately excluded

- **`axios` and `uuid`** — direct deps, already covered by open bot PRs: [#1251](https://github.com/4gray/iptvnator/pull/1251) bumps axios 1.16.0 → 1.18.1 (20 alerts), [#1252](https://github.com/4gray/iptvnator/pull/1252) covers uuid. Hand-bumping them here would only produce lockfile conflicts. Worth noting the open `uuid` CVE affects `v3/v5/v6` when `buf` is passed; this codebase uses only `v4` without `buf`, so it is unreachable — the 9 → 14 major is hygiene, not urgency.
- **`undici`** (7 advisories) — labelled `runtime` scope by Dependabot, but every path to it is build tooling: `electron` → `@electron/get`, `@angular/build`, `@module-federation/dts-plugin`. Not in the packaged app. Bumping it inside the Angular/Nx toolchain risks the build for no runtime benefit.

## Validation

| Check | Result |
| --- | --- |
| `nx test electron-backend` (`--skip-nx-cache`) | **75 suites / 851 tests passed** — covers `validated-axios`, `xtream.events`, `playlist.events` (form-data chain) and `app-update.events` (electron-updater/js-yaml chain) |
| `nx run-many --target=test` — `m3u-utils`, `ui-playback`, `m3u-state`, `playlist-m3u-feature-player` | passed |
| `nx build web` | passed — exercises `ajv@8.18.0` through `@angular-devkit/core` + `schema-utils`, the riskiest override |
| DASH ClearKey e2e (`apps/web-e2e/src/dash-clearkey.e2e.ts`) | **2 passed** (chromium); firefox/webkit skipped as usual, no ClearKey EME |

Each of the five chains was verified to resolve on disk, not just in the lockfile — pnpm leaves the superseded package block behind even when nothing depends on it, so grepping the lock is misleading:

```
mpd-parser       -> xmldom:    0.8.13
axios@1.16.0     -> form-data: 4.0.6
electron-updater -> js-yaml:   4.3.0
electron-conf    -> ajv:       8.18.0
ajv@8.18.0       -> fast-uri:  3.1.4
```

No test changes: this is a dependency bump with no behaviour change, so existing suites are the regression net. Docs updated — new `docs/architecture/dependency-security-overrides.md`.


4gray:
@codex review

Both earlier findings were addressed in `c7e46f88`, which landed after your review of `b9144e7d` (that commit is now `d2316faf` after a rebase onto master):

- **P1** — added `.changes/deps-transitive-cve-overrides.md`, typed `internal` per `.changes/README.md`, which names dependency bumps with behaviour risk as exactly that category. The `Release note gate` check now passes.
- **P2** — you were right that pnpm does not leave superseded package blocks behind. The section now describes the actual behaviour: the old versions survive only as the overrides' own selector keys in the `overrides:` block at the top of `pnpm-lock.yaml`, which is what a naive grep hits. The advice (resolve on disk, don't grep) is unchanged; the reasoning is now correct.

Please review `c7e46f88` — the docs commit has not been reviewed yet.

PR #1259, opened 2026-07-25
feat(release): manifest-driven screenshot capture with fail-closed mock-data guards

4gray:
> **Stacked on #1257** (release-note CI gate). GitHub will retarget this to `master` automatically once #1257 merges; the diff to review here is the single screenshots commit.

## Why

Release screenshots were single-use and fail-open. `capture-v020-screenshots.ts` was hard-coded to v0.20, and every isolation guarantee was hope, not proof:

| # | Hole | Consequence |
| --- | --- | --- |
| 1 | `getIptvnatorDataRoot()` silently falls back to `~/.iptvnator` if `IPTVNATOR_E2E_DATA_DIR` is lost | captures the user's **real playlists**, no error |
| 2 | `env: { ...process.env }` | ambient TMDB keys/proxies leak into a run that must be hermetic |
| 3 | no network gate | TMDB posters, logo CDNs, real streams can render into frames |
| 4 | only PNG existence checked | frame contents never validated |

## What

**Manifest-driven capture**: `screenshots.manifest.json` (slug → named setup steps → themes) + `capture-release-screenshots.ts`. Output: `apps/website/public/blog/<release>/screenshots/<slug>-<theme>.png` — exactly the paths the blog scaffold from #1256 emits. New feature shot = one manifest entry. `.changes` validation now cross-checks `screenshot:` slugs against the manifest.

**Fail-closed guards** (`screenshot-guards.mjs`, pure + unit-tested; any violation deletes the run's frames and exits non-zero):

- **G1** — real DB snapshotted (sha256+mtime) before launch, must be byte-identical after; the isolated DB must actually have been created
- **G2** — constructed env allowlist, never `...process.env`
- **G3** — deny-by-default network gate; the app's GitHub update check is answered by a **local stub** (deterministic, no network) — every other blocked request **fails the run**, because a silently-blocked TMDB call leaves a frame that looks broken instead of unsafe
- **G4** — pre-capture frame scan: external img/background URLs, credential-shaped text, MAC addresses, non-localhost `.m3u8` references
- **G5** — TMDB enrichment asserted disabled via the renderer's IndexedDB

## The guards worked on day one

The first live run was failed by G3: the xtream-mock-server **redirects stream endpoints to a public demo HLS** (`test-streams.mux.dev`). Earlier hand-run captures could therefore embed third-party video frames — precisely the copyright leak this PR exists to prevent. The M3U shot now captures the groups layout without starting playback.

The live runs also surfaced three selector drifts since v0.20 (add-playlist dialog is now a radio group, season cards became tabs with auto-select, portal entry links) — all fixed with order-independent navigation actions.

## Test plan

| Check | Result |
| --- | --- |
| `pnpm nx test release-tools` | 67/67 (20 new guard cases: manifest validation, env allowlist, URL gate incl. suffix-spoofed hostnames, stub exactness, frame report, DB snapshot/compare incl. bare-touch) |
| Live capture, dist build + mock server | 10/10 shots (5 slugs × dark/light) |
| Frames visually inspected | fictional titles/artwork only, no tooltips, no external resources |
| Guard failure paths | exercised live: G3 fired on mux.dev + api.github.com before the fixes |
| `pnpm nx lint release-tools`, `tsc --strict` | clean; all new TS files within the repo size limit |

No release note: tooling, `no-release-note` by its own policy.

### DIFF
--- a/.changes/README.md
+++ b/.changes/README.md
@@ -30,7 +30,7 @@ of the current season, watch progress, and click-to-play inline.
 | `type`       | yes      | `breaking`, `feature`, `fix`, `perf`, or `internal`          |
 | `area`       | yes      | lowercase slug, same as the conventional-commit scope        |
 | `issues`     | no       | issue numbers this closes — `[1187]` or `1187`               |
-| `screenshot` | no       | slug from the release screenshot manifest                    |
+| `screenshot` | no       | slug from `tools/release/screenshots.manifest.json`          |
 
 There is **no version field**. The release version is chosen deliberately at
 release time, not derived from these files.
@@ -84,3 +84,26 @@ The website publishes **one post per minor version** (`v0-18` … `v0-22`), and
 release screenshots live under the matching `blog/v0-24/` directory. A patch
 release therefore edits the existing post rather than generating a new one, so
 `--format blog` refuses to overwrite unless you pass `--force`.
+
+## Screenshots
+
+`pnpm run release:screenshots` captures every manifest shot in dark and light
+against the built app plus the Xtream mock server — never a real account.
+The run is fail-closed: it proves the real `~/.iptvnator/databases` directory
+(including the SQLite WAL sidecars, checked after Electron exits) was not
+touched, launches the app with an allowlisted environment, records and blocks
+all non-localhost traffic, scans every frame for external resources and
+credential-shaped text, and asserts TMDB enrichment stays disabled. Frames are
+staged outside the repository and published only once every shot and every
+guard has passed.
+
+Adding a shot for a new feature = one entry in
+`tools/release/screenshots.manifest.json` (plus, if navigation is new, one
+named action in `tools/release/capture-navigation.ts`).
+
+```bash
+pnpm nx run electron-backend:build-e2e   # once, before capturing
+pnpm run release:screenshots             # all shots, both themes
+pnpm run release:screenshots -- --only dashboard --theme dark
+pnpm run release:screenshots -- --release v0-24
+```
--- a/.claude/skills/release-cut/SKILL.md
+++ b/.claude/skills/release-cut/SKILL.md
@@ -42,11 +42,21 @@ is not optional.
     for a patch release, edit the existing post (the scaffold refuses to
     overwrite without `--force`).
 
-5. **Screenshots** — only from the release capture script against the mock
-   servers (`tools/release/`), never from a real playlist or account: real
-   streams, logos, and TMDB artwork are copyrighted, and credentials must
-   never reach a published image. Output goes to
-   `apps/website/public/blog/v0-XX/screenshots/`.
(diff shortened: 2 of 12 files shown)


4gray:
All three findings accepted and fixed in 876899a8 — and chasing the WAL one uncovered something worse.

**Codex P1 — network gate installed too late.** Correct: `page.route()` was registered after `electron.launch()` and after `findMainWindow()`'s 1.5 s settle, so startup traffic escaped both interception and `blockedRequests`. There is now a **main-process recorder** installed immediately after launch, before the renderer boots, which records every URL the Electron session sees; the page-level route still does the blocking. The verdict is computed over both lists by one shared function, so the layers cannot disagree about what counts as external.

**Greptile P1 — failure cleanup deleted prior screenshots.** Also correct: re-capturing over committed assets and failing on a later shot removed the freshly written files, leaving the originals already overwritten. Frames now land in a **staging directory outside the repository** and are moved into `blog/<release>/screenshots/` only after every shot and every guard has passed. A failure now leaves the published assets completely untouched.

**Codex P1 — WAL sidecars, and comparison before Electron exits.** Correct on both counts: `journal_mode = WAL` is set in `libs/shared/database/src/lib/connection.ts:994`, and my check snapshotted only `iptvnator.db` while the app was still running. G1 now snapshots the whole `databases/` directory including `-wal`/`-shm`, and the comparison runs in the `finally` **after** `app.close()`, once the DB worker has checkpointed.

### The finding behind the finding

Investigating the WAL comment exposed that **G1 had never actually run**. `snapshotFileState()` did `readFileSync()` + sha256; the production database here is **3.9 GB**, that read throws, and my own `catch` returned `{ exists: false }` — so every run compared "absent" to "absent" and reported success. Nine green runs, guard silently disabled. Exactly the fail-open class this PR exists to eliminate.

Identity is now size + mtime + inode per file (no hashing of multi-gigabyte data), only `ENOENT` counts as absent, and any other error propagates. Verified against the real database: 3 files seen including a 37 MB WAL.

I applied the same lesson to the new recorder: a recorder that observes nothing is indistinguishable from a clean run, so an **empty log now fails the run**. The live run reports `111 distinct request(s) observed, all local`, which proves the hook is wired rather than decorative.

### Verification

- 73/73 unit tests (G1 rewritten around directory snapshots: WAL-only writes, bare touch, added/removed sidecars, plus a regression test asserting no hashing)
- full live capture green: 10/10 shots, staging published only at the end, no leftovers
- lint + `tsc --strict` clean

4gray:
@greptileai review

4gray:
@codex review

4gray:
Superseded by #1261 — GitHub auto-closed this PR when its base branch was deleted after #1257 merged, and it could not be reopened because the head had been rebased onto master. All review feedback from this thread is applied there, with a summary table of every finding and its resolution.

PR #1261, opened 2026-07-25
feat(release): manifest-driven screenshot capture with fail-closed mock-data guards

4gray:
Replaces #1259, which GitHub auto-closed when its base branch was deleted after #1257 merged (the branch was rebased onto master, so it could not be reopened). Same work, all review feedback from that PR applied — summary of the rounds below.

## Why

`capture-v020-screenshots.ts` was single-use (hard-coded to v0.20) and every isolation guarantee was fail-open: a lost `IPTVNATOR_E2E_DATA_DIR` silently fell back to the user's real `~/.iptvnator` database, `...process.env` leaked ambient TMDB keys, nothing gated the network, and only PNG existence was checked. Each hole leaks real playlists, credentials, or copyrighted artwork into published screenshots with no signal.

## What

Manifest-driven capture (`screenshots.manifest.json` → slug, named setup steps, themes) writing `apps/website/public/blog/<release>/screenshots/<slug>-<theme>.png` — exactly the paths the blog scaffold from #1256 emits. `.changes` validation cross-checks `screenshot:` slugs against the manifest.

Fail-closed guards in `screenshot-guards.mjs` (pure, unit-tested):

- **G1** — the real `databases/` directory, **including SQLite WAL sidecars**, snapshotted before launch and compared **after Electron exits and checkpoints**
- **G2** — constructed env allowlist, never `...process.env`
- **G3** — main-process recorder installed before the renderer boots, plus a page-level deny-by-default route; any external attempt fails the run
- **G4** — pre-capture frame scan: external resources, credential-shaped text, MAC addresses, non-localhost `.m3u8`
- **G5** — TMDB asserted disabled via the renderer's IndexedDB

## Review rounds (all findings fixed)

| Finding | Resolution |
| --- | --- |
| Codex P1: network gate installed too late | main-process recorder before renderer boot; assertion now requires it to have seen the renderer's **own document load**, not merely "some" traffic |
| Codex P1: WAL sidecars, comparison before Electron exits | whole-directory snapshot incl. `-wal`/`-shm`, compared after `app.close()` |
| Codex P2: G1 skipped when the run throws | error captured, G1 evaluated on **every** exit path, G1 violation takes precedence and names the primary error |
| Greptile P1: cleanup deleted prior screenshots | frames staged outside the repo, published only after every shot and guard passes |
| Greptile P1: publication left a mixed set | copy-into-sibling + atomic directory swap with rollback (also fixes a latent `EXDEV` crash — staging is on a different filesystem) |

### Two bugs found by verifying, not by review

1. **G1 had never actually run.** `snapshotFileState()` did `readFileSync` + sha256; the production database here is **3.9 GB**, that read throws, and the `catch` returned `{ exists: false }` — nine green runs compared "absent" to "absent". Now size + mtime + inode per file, `ENOENT` distinguished from real errors. Verified against the real DB: 3 files seen including a 37 MB WAL. The same lesson was applied to the recorder — an empty log now fails the run.
2. **The fix for the mixed-set finding introduced a worse loss.** Replacing the directory wholesale meant `--only dashboard` deleted the other 8 screenshots — the exact flag used to refresh one shot before a release. Publication now merges for filtered runs and replaces for full runs.

## Test plan

| Check | Result |
| --- | --- |
| `pnpm nx test release-tools` | 79/79 |
| Live capture | 10/10 shots; `Guards passed: 111 distinct request(s) observed, all local` |
| Filtered republish | `--only dashboard` → directory still holds 10 files, `dashboard-*` updated |
| Frames inspected | fictional titles/artwork only |
| Guard failure paths | exercised live (G3 caught the mock server redirecting streams to `test-streams.mux.dev`; the M3U shot no longer starts playback) |
| lint + `tsc --strict` | clean; all files within the repo size limit |

No release note: tooling, `no-release-note` by its own policy.

### DIFF
--- a/.changes/README.md
+++ b/.changes/README.md
@@ -30,7 +30,7 @@ of the current season, watch progress, and click-to-play inline.
 | `type`       | yes      | `breaking`, `feature`, `fix`, `perf`, or `internal`          |
 | `area`       | yes      | lowercase slug, same as the conventional-commit scope        |
 | `issues`     | no       | issue numbers this closes — `[1187]` or `1187`               |
-| `screenshot` | no       | slug from the release screenshot manifest                    |
+| `screenshot` | no       | slug from `tools/release/screenshots.manifest.json`          |
 
 There is **no version field**. The release version is chosen deliberately at
 release time, not derived from these files.
@@ -84,3 +84,26 @@ The website publishes **one post per minor version** (`v0-18` … `v0-22`), and
 release screenshots live under the matching `blog/v0-24/` directory. A patch
 release therefore edits the existing post rather than generating a new one, so
 `--format blog` refuses to overwrite unless you pass `--force`.
+
+## Screenshots
+
+`pnpm run release:screenshots` captures every manifest shot in dark and light
+against the built app plus the Xtream mock server — never a real account.
+The run is fail-closed: it proves the real `~/.iptvnator/databases` directory
+(including the SQLite WAL sidecars, checked after Electron exits) was not
+touched, launches the app with an allowlisted environment, records and blocks
+all non-localhost traffic, scans every frame for external resources and
+credential-shaped text, and asserts TMDB enrichment stays disabled. Frames are
+staged outside the repository and published only once every shot and every
+guard has passed.
+
+Adding a shot for a new feature = one entry in
+`tools/release/screenshots.manifest.json` (plus, if navigation is new, one
+named action in `tools/release/capture-navigation.ts`).
+
+```bash
+pnpm nx run electron-backend:build-e2e   # once, before capturing
+pnpm run release:screenshots             # all shots, both themes
+pnpm run release:screenshots -- --only dashboard --theme dark
+pnpm run release:screenshots -- --release v0-24
+```
--- a/.claude/skills/release-cut/SKILL.md
+++ b/.claude/skills/release-cut/SKILL.md
@@ -42,11 +42,21 @@ is not optional.
     for a patch release, edit the existing post (the scaffold refuses to
     overwrite without `--force`).
 
-5. **Screenshots** — only from the release capture script against the mock
-   servers (`tools/release/`), never from a real playlist or account: real
-   streams, logos, and TMDB artwork are copyrighted, and credentials must
-   never reach a published image. Output goes to
-   `apps/website/public/blog/v0-XX/screenshots/`.
+5. **Screenshots** — only from the fail-closed capture script against the
+   mock servers, never from a real playlist or account: real streams, logos,
+   and TMDB artwork are copyrighted, and credentials must never reach a
+   published image.
+
+    ```bash
+    pnpm nx run electron-backend:build-e2e   # once
+    pnpm run release:screenshots             # all manifest shots, dark+light
+    ```
+
+    Output goes to `apps/website/public/blog/v0-XX/screenshots/`. New feature
+    to showcase = new entry in `tools/release/screenshots.manifest.json`
+    (slug must match the note's `screenshot:` field). The run aborts and
+    deletes its frames on any guard violation (real-DB touch, external
+    request, credential-shaped text in frame, TMDB active).
 
 6. **Consume the notes** (the only destructive step):
 
--- a/.codex/skills/release-cut/SKILL.md
+++ b/.codex/skills/release-cut/SKILL.md
@@ -42,11 +42,21 @@ is not optional.
     for a patch release, edit the existing post (the scaffold refuses to
     overwrite without `--force`).
 
-5. **Screenshots** — only from the release capture script against the mock
-   servers (`tools/release/`), never from a real playlist or account: real
-   streams, logos, and TMDB artwork are copyrighted, and credentials must
-   never reach a published image. Output goes to
-   `apps/website/public/blog/v0-XX/screenshots/`.
+5. **Screenshots** — only from the fail-closed capture script against the
+   mock servers, never from a real playlist or account: real streams, logos,
+   and TMDB artwork are copyrighted, and credentials must never reach a
+   published image.
+
+    ```bash
+    pnpm nx run electron-backend:build-e2e   # once
+    pnpm run release:screenshots             # all manifest shots, dark+light
+    ```
+
+    Output goes to `apps/website/public/blog/v0-XX/screenshots/`. New feature
+    to showcase = new entry in `tools/release/screenshots.manifest.json`
+    (slug must match the note's `screenshot:` field). The run aborts and
+    deletes its frames on any guard violation (real-DB touch, external
+    request, credential-shaped text in frame, TMDB active).
 
 6. **Consume the notes** (the only destructive step):
 
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
         "release:notes:github": "node tools/release/build-release-notes.mjs --format github",
         "release:notes:changelog": "node tools/release/build-release-notes.mjs --format changelog",
         "release:notes:blog": "node tools/release/build-release-notes.mjs --format blog",
+        "release:screenshots": "tsx tools/release/capture-release-screenshots.ts",
         "lint": "nx run-many --target=lint --all",
         "build": "nx build electron-backend"
     },
--- a/tools/release/build-release-notes.mjs
+++ b/tools/release/build-release-notes.mjs
@@ -19,6 +19,7 @@ import process from 'node:process';
 import { fileURLToPath } from 'node:url';
 
 import { loadNotes } from './release-notes.mjs';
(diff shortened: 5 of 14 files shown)


4gray:
All four findings fixed in 8c064ed2 — and the live verification caught a fifth that no static check could.

**CodeQL high (`js/incomplete-sanitization`)** — `clickHrefSuffix` built a `RegExp` from a string while escaping only slashes. Removed the regex entirely: `waitForURL` accepts a predicate, so no escaping is needed. Also replaced the last dynamic `RegExp` in the driver with literal patterns.

**Greptile P1 — main-process traffic remained unblocked.** The important one, and correct: the session hook only recorded (`cancel: false`) while `page.route` covers renderer traffic exclusively, so anything the main process fetched would genuinely reach the network and only be reported afterwards. The hook now **blocks**: local schemes/hosts and the exact stub prefixes pass, everything else is cancelled. The policy data is single-sourced through `networkPolicy()` (the predicate is written twice — once in-process, once serialized into Electron main — so a test asserts the two agree).

**Greptile P1 — `--release` path traversal.** Real and nasty in combination with replace mode, which renames and recursively deletes the resolved directory. `--release` is now validated as a slug, plus an independent assertion that the resolved output path stays inside `apps/website/public/blog/`.

**Greptile P1 — invalid `--theme`.** `as Theme` erased the check, so `--theme --only` was accepted and every non-`dark` value rendered as light into a file like `---only.png`. The value is now checked against the manifest's themes.

### The fifth bug, found by running it

The blocking hook failed at install with `ReferenceError: __name is not defined`. `app.evaluate()` serializes its body into the Electron main process, and esbuild (via tsx) wraps **named** functions with a `__name` helper that does not exist there. The previous version survived only because it contained nothing but anonymous callbacks; my `const isLocal = …` broke it. Rewritten inline, with a comment so it is not reintroduced.

Worth stating plainly: 83 green tests, clean `tsc --strict` and lint said nothing about whether a closure survives crossing a process boundary. Only the live run did. The failure was at least loud — the recorder throws, the run aborts, nothing is published — so the fail-closed design held.

### Verification

- 83/83 unit tests; lint + `tsc --strict` clean
- full live capture: 10/10 shots, `Guards passed: 111 distinct request(s) observed, all local`
- frames inspected: fictional titles/artwork only
- all three new argument guards exercised live: traversal, bad theme, and unknown `--only` slug each fail fast with an explicit message

Follow-up worth a separate PR: `tools/release/capture-v020-screenshots.ts` is fully superseded by this script and still carries the same unsafe `RegExp` construction CodeQL flagged here. Deleting it would remove ~1165 lines and silence those alerts, but it is out of scope for this diff.

4gray:
@greptileai review

4gray:
Both remaining findings fixed in 878a4968.

**Codex P1 — recorder installed after Electron resumes startup.** The objection lands, and it specifically defeats my previous "strengthening": seeing the `index.html` request proves the hook existed before *that* request, not before an earlier one. Playwright offers no pre-ready hook, so I closed the window from outside the app instead — the capture now launches Electron with

```
--host-resolver-rules=MAP * ~NOTFOUND, EXCLUDE localhost, EXCLUDE 127.0.0.1
```

which takes effect before Electron runs a line of app code. There is no install-timing window at all.

Stated precisely rather than overclaimed: the switch covers the **Chromium** stack (renderer plus the main-process `net` module). Requests made through Node's own stack from main — axios, for instance — are not affected by it, which is exactly why the session hook, the page route and the recorded-URL verdict all still exist. That limitation is written into the constant's doc comment, not glossed over. In practice a capture run makes no Node-stack requests: the only import is from the local fixture.

**Codex P2 — manifest themes unvalidated.** Correct and worth more than it looks: `["dark"]` or a typo like `["drak"]` passed validation, a full run publishes in `replace` mode and deletes the previous set, and the blog scaffold references both `-dark.png` and `-light.png` — the release post ends up with broken images. The manifest must now carry exactly `dark` and `light` (order-independent), asserted against the committed manifest too.

The network gate is now four layers, and only the first is timing-independent:

1. `--host-resolver-rules` — before any app code, Chromium stack
2. session `webRequest` hook — records and blocks
3. `page.route` — renderer, also fulfills the stubbed update check locally
4. G4 frame scan + the G3 verdict over every recorded URL — the run fails if anything external was even attempted

### Verification

- 86/86 unit tests; lint + `tsc --strict` clean
- full live capture with the resolver gate active: 10/10 shots, `Guards passed: 111 distinct request(s) observed, all local` — the new switch does not starve the app of the local mock
- the committed manifest passes the stricter validator

4gray:
@greptileai review

4gray:
@codex review

4gray:
Fixed in d561ed8c. One correction on the blocking claim, and the underlying concern closed anyway.

### On the stated blocker

> the startup update path can still send an external GitHub request through Node fetch

I checked rather than assuming, and the specific path does not fire in a capture run:

- `checkForUpdatesOnStartup()` returns early on `!this.isPackaged` (`app-update.service.ts:316`), and the capture launches an unpackaged `dist/apps/electron-backend/main.js`
- the other `fetch` user, `getReleaseNotes`, is only reachable from the release-notes dialog, which no manifest shot opens

So no such request leaves today. **The architectural point stands regardless**, and it is the more valuable half: main-process `fetch` runs on Node's stack, invisible to the resolver switch, the `webRequest` hook and `page.route` alike — so a future call added anywhere in main would pierce the gate silently.

Closed at the stack level rather than for that one call site: `globalThis.fetch` in the main process is now wrapped to record every URL into the same log and reject non-local ones. That also means the recorded-URL verdict covers it, so a regression fails the run instead of shipping a frame.

### Codex P1 — mock server identity

Sharp catch, and it undercut a real assumption: every guard treats localhost as trusted, so any 2xx on `/health` was enough to hand an unknown local server the role of fixture source, with its data going straight into published screenshots. The script now requires the service to answer the Xtream marketing API with the synthetic categories these shots are built around, and refuses to capture otherwise.

### Codex P2 — XAUTHORITY

Correct: on xauth-based X11 sessions, stripping it stops Electron from opening the display, so the documented command dies before a window exists. Added to the platform allowlist with a note that it grants display access, not application credentials.

The gate now has five layers, in order of when they take effect:

1. `--host-resolver-rules` — before any app code (Chromium stack)
2. wrapped main-process `fetch` — Node stack
3. session `webRequest` hook — records and blocks
4. `page.route` — renderer, fulfills the stubbed update check locally
5. G4 frame scan + G3 verdict over every recorded URL

Files were split to stay under the 400-line rule: `capture-network-gate.ts` and `capture-tmdb-check.ts` are now their own modules (358 / 361 / 250 / 107 / 56 lines).

### Verification

- 87/87 unit tests; lint + `tsc --strict` clean
- full live capture: 10/10 shots, `Guards passed: 111 distinct request(s) observed, all local` — the `fetch` wrapper and the identity check do not disturb the run

4gray:
@greptileai review

4gray:
@codex review

4gray:
I tried to close this one properly and could not — here is the evidence, so the residual window is documented rather than glossed over.

**The attempt.** A `NODE_OPTIONS=--require` preload runs in Electron's main process *before* `app.whenReady()`, which would have removed the window entirely. Verified standalone: `NODE_OPTIONS="--require pre.cjs" electron probe.js` → the preload runs first.

**Why it cannot work here.** Under Playwright it never arrives — `process.env.NODE_OPTIONS` is `null` inside the launched app. The reason is unconditional, in `playwright-core/lib/server/electron/electron.js:164`:

```js
delete env.NODE_OPTIONS;
```

Playwright also exposes no pre-ready hook and no `connectOverCDP` path for Electron, so there is no test-side seam left. The remaining option is your second suggestion — holding renderer startup from inside `main.ts` — and I am not willing to bend production startup code around a screenshot tool.

Worth noting the attempt was not wasted: the self-check I added (`preload gate did not run …`) is what caught it. The gap announced itself loudly instead of passing as a green run — which is the property this whole PR is about.

**What actually covers that window**, now written into the module's doc comment rather than claimed as complete coverage:

- Chromium-stack traffic — renderer and main-process `net` — is blocked from process start by `--host-resolver-rules`, which has no timing window at all
- the only startup Node `fetch` in the app is the update check, and it returns early on `!app.isPackaged` (`app-update.service.ts:316`); the capture always runs unpackaged
- anything that does slip through is still recorded and fails the run's G3 verdict, so a regression cannot ship quietly

So the exposure is: a Node-stack request, from the main process, in the first milliseconds after ready, that does not exist today and would fail the run if introduced. I judged that acceptable and, more to the point, unclosable from where this code sits. Happy to revisit if you see a seam I missed.

87/87 tests, lint and `tsc --strict` clean, reverted to the verified state (`d561ed8c` content) with only the documentation added.

PR #1266, opened 2026-07-26
refactor(deps): drop uuid for a shared crypto-based id helper

4gray:
Replaces #1252 (and the stale #872).

## Why the bump was not just a lockfile change

- **uuid 14 is ESM-only** (`"type": "module"`). That is what produced the wall of `Jest encountered an unexpected token` on #872, and it is not covered by the libs' `transformIgnorePatterns: ['node_modules/(?!.*\\.mjs$)']`.
- `apps/web/jest.config.ts` only kept uuid working by mapping `^uuid$` to `node_modules/uuid/wrapper.mjs` — a file v14 no longer ships.
- The specifier also has to be synced in `libs/shared/m3u-utils/package.json`, otherwise `@nx/dependency-checks` fails lint. That is the actual `Lint` error on #1252: *"The version specifier does not contain the installed version of "uuid" package: 14.0.1"*.
- `@types/uuid` would have to go anyway, since v14 ships its own types.

All four call sites only ever used `v4()`, so the dependency goes away instead of forward.

## The helper

`createRandomId()` in `@iptvnator/shared/interfaces` prefers `crypto.randomUUID()` and falls back to building the same v4 shape from `crypto.getRandomValues()`.

That fallback is load-bearing, not defensive padding: **`crypto.randomUUID` is only exposed in secure contexts**, and the self-hosted Docker PWA is regularly served over plain http on a LAN address, which is not one. `getRandomValues` stays available there — and it is exactly what uuid's own `v4()` used, so behaviour is unchanged on both paths.

## Validation

| command | result |
| --- | --- |
| `nx run-many --target=test --projects=shared-interfaces,m3u-utils,services,playlist-import-feature` | 4 projects, all pass (incl. 8 new `random-id.util.spec.ts` cases) |
| `nx test web` | 157 passed |
| `nx run-many --target=lint --projects=shared-interfaces,m3u-utils,services,playlist-import-feature,web` | 5 projects pass, 0 errors |
| `pnpm run typecheck:web` / `typecheck:backend` | clean |

The new spec covers the v4 format, uniqueness, native-`randomUUID` delegation, and the insecure-context path — including the version/variant bits on all-zero and all-`0xff` randomness.

`no-release-note`: no behaviour change, same v4 identifiers, so this is a pure refactor under `.changes/README.md`.


4gray:
Rebased onto master after #1270 landed (`b62aac86`).

Conflicts were only in `package.json` and `pnpm-lock.yaml`. Resolved by keeping master's bumped versions (shaka-player 5.2.2, video.js 8.23.9, typescript-eslint 8.65) and dropping `uuid` + `@types/uuid` on top; the lockfile was regenerated from master's, so its diff is now just the two removed entries (2 insertions, 21 deletions).

Re-verified after the rebase: `shared-interfaces`/`m3u-utils`/`services`/`playlist-import-feature` all pass, `nx test web` 157 passed, lint clean on all five projects, both typechecks clean.

PR #1269, opened 2026-07-26
fix(playback): make the "Show subtitles" setting reach the web players

### LINKED ISSUE #1155: Subtitles Preferences 0.22
**Describe the bug**
Preferences for subtitles have no effect

**To Reproduce**
switch Preferences for subtitles on / off, Subtitles are always displayed

**Expected behavior**
No subtitles when switched off

-   OS: LMDE7 (Debian 13)
-   iptvnator Version 0.22

4gray:
Closes #1155.

## The bug

`Settings.showCaptions` never had an owner outside the default-off shared web-controls flag, so with the shipping controls it was a no-op:

| Player | Behaviour before |
| --- | --- |
| Video.js (default) | input not read at all — in v0.22 the component did not even declare it |
| ArtPlayer | input declared, never used (except DASH/Shaka) |
| HTML5 | one-shot pass after `play()` resolved — runs before hls.js adds its text tracks, so an `AUTOSELECT=YES,DEFAULT=YES` rendition stayed on screen |

On top of that no portal host bound the input: Xtream live, Stalker live, the unified live tab and the portal detail inline player all left it at the `false` default, so only the M3U player passed it through at all.

## The fix

Extract the source-local track controllers into an adapter-free `WebVideoSourceTracks`; `WebVideoSourceControlsBridge` now wraps it and adds adapter attach/detach plus MPEG-TS VOD duration correction. Both controls modes go through the same caption code.

The preference-off paths bind it directly:

- **HTML5** and **ArtPlayer** feed `WebVideoSourceTracks` the same active source as their controls bridge (`bindControlsSource` / `prepareForSourceChange`).
- **Video.js** gets a new `VjsLegacyTracks` (audio + text) in place of the audio-only helper, bound at the same lifecycle points.
- `HtmlVideoElementSession` loses its caption concern; the `play()`-time one-shot is gone.

The preference is applied when a source binds and re-applied as the engine adds or switches tracks — which is what the old check could not do.

`WebPlayerViewComponent` reads the preference from `SettingsStore` instead of an input, so every host inherits it without wiring. The duplicated `playerSettings.showCaptions` plumbing in the M3U player is removed. Toggling the setting applies to the playing stream immediately (`updateSettings` patches the store → `ngOnChanges` → `refreshInputs()`).

## Caption ownership: two modes

The preference means different things depending on who renders the caption UI, and this is the part Codex caught in review:

- **Shared controls — authoritative.** They own the caption menu, so user intent arrives through `setSubtitleTrack` and the explicit override (including `-1` = off) wins until the source changes. Unchanged by this PR.
- **Vendor chrome — source-default.** The engine still renders its own caption menu, so the preference seeds each new source and is released once the media element reports `playing`. Engine defaults land well before that; a user opening the caption menu does not. Enforcing it session-long would have made that menu inert — and since the setting defaults to off, that would have broken the caption button for everyone who never opened settings.

Mode selection is an optional `playbackStarted` probe the legacy owners pass to the track helpers: `WebVideoSourceTracks` listens on the media element, `VjsLegacyTracks` on the Video.js player, both reset per source. Shared controls omit it.

On the HLS side, source-default mode deselects the track (`subtitleTrack = -1`) instead of setting `subtitleDisplay = false`: hls.js applies `subtitleDisplay` to whatever the vendor menu picks, so hiding would silently override the user, while `-1` also clears hls.js' own `selectDefaultTrack` so it cannot reselect a default later.

## Tests

Every new spec was verified to fail with the fix reverted:

| Spec | Covers | Fails without fix |
| --- | --- | --- |
| `html-video-player.component.captions.spec.ts` | initial + late track, preference on, user selection after `playing`, re-seed on next channel, restore on re-enable | 3 / 6 |
| `art-player-source-session.captions.spec.ts` | HLS default deselected, selection after `playing` kept, preference on, source replacement, teardown | 3 / 5 |
| `vjs-player.component.captions.spec.ts` | VHS text tracks, selection after `playing` kept, new source, destroy | 4 / 6 |
| `web-player-view.component.shared-controls.spec.ts` | preference reaches players with no host binding | — |

Stale specs for the removed one-shot behaviour were updated rather than deleted (`html-video-element-session`, `html-video-player.component`, `html-video-player.component.shared-controls`).

**No E2E added**: no existing E2E touches subtitles and the mock servers serve no subtitle renditions, so the behaviour is not reproducible at that level. Coverage sits at the engine boundary instead.

```
ui-playback              83 suites / 723 tests   pass
web                      20 suites / 157 tests   pass
playlist-m3u-feature-player  3 suites /  43 tests  pass
services                 26 suites / 245 tests   pass
portal-xtream-feature    17 suites / 144 tests   pass
portal-stalker-feature    5 suites /  67 tests   pass
portal-shared-ui          9 suites /  78 tests   pass
nx affected --target=lint                        9 projects clean
```

Three files had to be split to stay under the 400-line `max-lines` rule (`VjsLegacyTracks`, `web-player-view.spec-stubs.ts`, and the caption specs). Nothing was added to the baseline.

## Docs & release note

- `.changes/playback-subtitle-preference.md` (`type: fix`, `issues: [1155]`) — `release:notes:validate` passes
- `docs/architecture/player-controls-contract.md` — new "Caption preference in both modes" section, updated file map and `HtmlVideoElementSession` description
- `CLAUDE.md` + `AGENTS.md` kept in sync: `showCaptions` is explicitly outside the shared-controls rollout gate


4gray:
@greptileai

4gray:
@codex review

4gray:
@greptileai

4gray:
@codex review

PR #1278, opened 2026-07-26
refactor(electron-backend): split four files under the max-lines limit

4gray:
## Why

Four TypeScript files exceeded the 400-line `max-lines` hard limit and failed `eslint` on `master`:

| File | Lines |
|---|---|
| `embedded-mpv-frame-copy.adapter.spec.ts` | 547 |
| `embedded-mpv-linux-linkage.spec.ts` | 539 |
| `epg.events.ts` | 514 |
| `embedded-mpv-frame-copy.adapter.ts` | 428 |

Committing a regenerated baseline would have been the wrong fix — the baseline header says the list must only shrink.

## One correction worth flagging

A regenerated baseline also wanted to add four `apps/electron-backend-e2e/src/performance/*` files. Those **do not error**: each already carries a file-wide `/* eslint-disable max-lines -- <why> */` (e.g. `m3u-refresh-main-capture.ts` is serialized into the Electron main process by `electronApp.evaluate` and genuinely cannot be split).

The generator is a pure line counter and can't see inline disables, so it wanted to re-add them on every run — which is *why* the baseline could never "only shrink". Fixed here: the generator now skips files with a justified file-wide disable, matching the escape hatch already documented for exactly this case.

## What changed

**Splits** (all now 266–298 lines, under the <300 target):

- `epg.events.ts` → keeps IPC registration; `epg-fetch.service.ts` (freshness + fetch orchestration), `epg-mapping.service.ts` (mapping resolution + CRUD)
- `embedded-mpv-frame-copy.adapter.ts` → keeps process lifecycle; `-protocol.ts` (wire protocol), `-spawn.ts` (launch/env resolution)
- both specs split by concern, shared fixtures in `*.test-helpers.ts` (matching the existing `runtime-fixtures.test-helpers.ts` convention)

**Baseline result:** file list is byte-identical — 128 entries before and after, zero added, zero removed. Only the banner gained two lines documenting the escape hatch. Rerunning the generator is idempotent.

## Behavior

No change. Error-handling semantics were preserved verbatim — including a quirk I deliberately did **not** fix: four EPG mapping handlers `return` their DB promise inside a `try/catch` instead of awaiting it, so the catch only covers `getDatabase()` failures while query rejections escape to the IPC caller. Tracked separately so this stays a pure refactor.

## Validation

- `pnpm nx lint electron-backend` — pass
- `pnpm nx lint electron-backend-e2e` — pass (40 pre-existing Playwright warnings, 0 errors)
- `pnpm nx test electron-backend` — **872 passed, 79 suites**; per-split counts held constant (EPG 14, frame-copy adapter 17, linkage 16)
- `pnpm run release:notes:validate` — 9 notes valid
- Generator matcher checked against 7 cases including the `max-lines-per-function` and `eslint-disable-next-line` near-misses

**Caveat:** verified on darwin only. The Linux linkage/frame-copy specs pass here because they stub `process.platform`; the Linux CI run is the real check.

## Docs

`CLAUDE.md` (max-lines section + `epg.events.ts` description) and `docs/architecture/sqlite-db-worker.md` (EPG ownership list, whose numbering was also off). `AGENTS.md` mirrors only the process sections, so it needed no change.

Release note intentionally skipped — pure refactor, hence the `no-release-note` label.


4gray:
Addressed in 9506ff55 — added the max-lines baseline workflow (list only shrinks; justified file-wide `eslint-disable max-lines` as the escape hatch) to the **Agent Bootstrap** section of `AGENTS.md`.

Worth noting the finding was slightly off on the letter but right on the substance: `AGENTS.md` had no linting section at all and never mentioned `max-lines`, so this was a pre-existing gap rather than drift introduced here, and "Linting" is not one of the sections `CLAUDE.md` lists as mirrored. The underlying risk is real though — and somewhat ironic — since an agent bootstrapping from `AGENTS.md` could regenerate the baseline and commit the four justified `performance/*` exemptions right back in, which is exactly what the generator change in this PR prevents. Closed it for that reason.

PR #1279, opened 2026-07-26
fix(epg): await mapping queries so lookups fail soft

4gray:
## Problem

Four EPG mapping handlers in `apps/electron-backend/src/app/events/epg-mapping.service.ts` wrap their DB call in `try`/`catch` but `return` the promise instead of `await`ing it:

```ts
try {
    const db = await getDatabase();
    return getEpgMapping(db, channelKey);   // adopted, not awaited
} catch {
    return null;
}
```

An `async` function **adopts** a returned promise rather than awaiting it, so the `catch` only ever fired when `getDatabase()` itself threw. A rejection from the underlying query (`getEpgMapping` / `setEpgMapping` / `deleteEpgMapping` / `searchEpgChannels`) propagated straight to the IPC caller instead of returning the intended `null` / `{success: false}` / `[]` fallback.

`handleGetEpgMappingsBatch` and `resolveChannelIds` in the same module already `await`, so they fail soft for both cases. The fallback values and the module's own doc comment — *"Every method fails soft: a mapping lookup must never take down an EPG request"* — show fail-soft is the intended contract throughout. These four just missed it.

Pre-existing, not a regression: the behavior predates the max-lines split in #1278, which carried it over verbatim from `epg.events.ts`.

## Fix

Added the missing `await` to `handleGetEpgMapping`, `handleSetEpgMapping`, `handleDeleteEpgMapping` and `handleSearchEpgChannels`. Four lines.

## Tests

New `apps/electron-backend/src/app/events/epg-mapping.events.spec.ts` (174 lines) drives the **real IPC handlers** captured from a mocked `ipcMain.handle`, so it asserts the contract that actually matters: the caller receives a fallback, not a rejected promise. It covers the `epg.events.ts` → `epg-mapping.service.ts` delegation as well as the handlers themselves.

17 tests — for each of the 5 mapping channels: a rejecting operation, a rejecting `getDatabase()`, and a success passthrough, plus the two short-circuit paths (blank search term, empty batch).

Verified the coverage binds to this bug: reverting the four `await`s produces **exactly 4 failures**, one per fixed handler, while `EPG_MAPPING_GET_BATCH` (which already awaited) stays green.

## Validation

| Command | Result |
| --- | --- |
| `pnpm nx test electron-backend` | 95 suites, 1027 passed |
| `pnpm nx lint electron-backend` | pass |
| `npx eslint` on all three changed files | clean |
| `pnpm run release:notes:validate` | valid |

## Note for the reviewer

Rebased onto `master` after #1278 landed; the fix moved from `epg.events.ts` into the newly extracted `epg-mapping.service.ts`, where the bug had been carried over unchanged.

One unrelated pre-existing issue surfaced while verifying, tracked separately and **not** addressed here: the `lint` target in `apps/electron-backend/project.json` uses an unquoted `apps/electron-backend/**/*.ts`. Nx shells out to `/bin/sh`, where `**` is not recursive, so it expands to just `src/main.ts` — the target lints 1 file of 222 and reports success. The two *quoted* globs in the same command work fine, since eslint then does its own globbing.

### DIFF
--- a/.changes/epg-mapping-lookup-fail-soft.md
+++ b/.changes/epg-mapping-lookup-fail-soft.md
@@ -1,0 +1,9 @@
+---
+type: fix
+area: epg
+---
+
+A database error while reading or saving a manual EPG channel mapping no longer
+surfaces as a failed request. Looking up, saving, deleting, and searching
+mappings now fall back quietly, so a transient storage hiccup can no longer take
+down the EPG panel or the "Map EPG channel" dialog.
--- a/apps/electron-backend/src/app/events/epg-mapping.events.spec.ts
+++ b/apps/electron-backend/src/app/events/epg-mapping.events.spec.ts
@@ -1,0 +1,174 @@
+import type EpgEventsType from './epg.events';
+
+const getDatabase = jest.fn();
+const getEpgMapping = jest.fn();
+const getEpgMappingsBatch = jest.fn();
+const setEpgMapping = jest.fn();
+const deleteEpgMapping = jest.fn();
+const searchEpgChannels = jest.fn();
+const ipcHandlers = new Map<
+    string,
+    (event: unknown, args: unknown) => Promise<unknown>
+>();
+
+jest.mock('electron', () => ({
+    app: {
+        isPackaged: false,
+        getAppPath: () => '/mock/app.asar',
+    },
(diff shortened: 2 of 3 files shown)


4gray:
@greptileai

PR #1288, opened 2026-07-27
refactor(electron-backend): split the last four files over the max-lines cap

4gray:
Follow-up to #1278. That PR split four of the files the `electron-backend` lint target had been silently skipping; four were left over. After this, no file in the project sits above the 400-line cap outside the baseline.

### Why these were invisible

The lint target's first glob is unquoted:

```
eslint apps/electron-backend/**/*.ts "…frame-copy-runtime.ts" "…frame-copy-runtime/**/*.ts"
```

Nx runs commands through `/bin/sh`, which has no `globstar`, so `apps/electron-backend/**/*.ts` collapses to `src/main.ts` — the target lints **1 file instead of 245** and reports success. The quoting fix belongs to #1176 and is deliberately **not** in this PR, so the two land without overlapping.

### What changed

Nothing is added to `tools/eslint/max-lines-baseline.mjs` — the baseline only shrinks. Shared setup moves to `*.test-helpers.ts`, the suffix `tsconfig.app.json` already excludes, so the production build never sees jest globals.

| File | Before | After | New file |
|---|---|---|---|
| `remote-control.events.spec.ts` | 588 | 188 | `remote-control-http.spec.ts` |
| `downloads.events.spec.ts` | 530 | 182 | `downloads-actions.spec.ts` |
| `http-server.spec.ts` | 448 | 377 | `resolve-static-file-path.spec.ts` |
| `worker-performance-capture.spec.ts` | 408 | 149 | `worker-performance-capture.resilience.spec.ts` |

- **remote-control** — the mock registry moves to `remote-control.test-helpers.ts`; each spec keeps its own `jest.mock()` factories, which resolve the `mock`-prefixed exports.
- **downloads** — `jest.doMock` is not hoisted, so the whole harness moves behind `setupDownloadsEventsHarness()`.
- **http-server** — the `resolveStaticFilePath` cases were already self-contained.
- **worker-performance-capture** — named `.resilience.spec.ts` to match the existing `.concurrency` / `.histogram` siblings.

Test bodies are unchanged; the helpers keep the same identifiers in scope, so these are moves, not rewrites.

### Verification

With the glob quoted locally (again — that quoting is #1176's, not part of this PR):

- **245 files linted, 0 `max-lines` errors.** The only remaining lint errors are the five #1176 fixes (`prefer-const` ×2, `no-unsafe-finally` ×2, `no-control-regex`).
- `tsconfig.app.json` and `tsconfig.spec.json` both typecheck clean.
- Full suite: 1045 tests, 103 suites. Per-file test counts unchanged across every split.

### Pre-existing flake, not from this PR

`worker-performance-capture.concurrency.spec.ts` › *"reliably observes a real 20ms event-loop block after condition-based arming"* fails intermittently. It asserts on a real measured event-loop block, so it is sensitive to machine load. **It failed 4/4 full-suite runs on a clean checkout of master** (24f0dee6, no local changes) — it is not caused by this change. Introduced in da3b6572.

### Release note

Skipped — pure refactor, no user-visible behavior change. Labelled `no-release-note`.


4gray:
@codex review

PR #1291, opened 2026-07-27
fix(epg): make manual EPG mapping lookups actually fail soft

4gray:
## What

Four EPG mapping handlers documented a fail-soft contract but only honored half of it. They returned the database operation's promise from inside the `try` block **without awaiting it**, so the rejection escaped the `catch` — the try block exits before the promise settles:

```ts
try {
    const db = await getDatabase();
    return getEpgMapping(db, channelKey);  // rejection escapes the catch
} catch {
    return null;
}
```

A `getDatabase()` failure *was* caught, but a SQLite error inside the operation rejected the `ipcMain.handle` promise, so the renderer call threw instead of receiving `null` / `{success:false}` / `[]`.

User-visible effect: a database error while searching in the **"Map EPG channel"** dialog, or while saving/removing a mapping, failed the request outright instead of degrading to an empty result.

Adding `await` in `handleGetEpgMapping`, `handleSetEpgMapping`, `handleDeleteEpgMapping` and `handleSearchEpgChannels` closes it. `resolveChannelIds` and `handleGetEpgMappingsBatch` already awaited correctly and are unchanged.

## Not a regression

The identical shape predates the `epg.events.ts` split in 636545cb (#1278) — that refactor preserved semantics faithfully and inherited the bug. Confirmed via `git show 636545cb^`.

## Tests

`epg-mapping.service.spec.ts` is new — these handlers had **no coverage at all** before this PR. Table-driven across all six functions × both failure modes (rejecting `getDatabase()` **and** rejecting operation), plus the guard clauses, the `Map`→record conversion, and `queryByResolvedChannelIds` remapping / identity fallback.

Verified the spec actually fails on the old behavior rather than assuming it would: reverting the four `await`s fails **exactly** the four operation-rejection cases (4 failed, 21 passed), then restored.

| Check | Result |
|---|---|
| `pnpm nx test electron-backend` | 95 suites, 1035 tests passed |
| New spec | 25 passed |
| Regression proof (fix reverted) | 4 failed — exactly the expected cases |
| `npx eslint` on changed files | 0 errors, 0 warnings |
| `npx prettier --check` | clean |
| `pnpm run release:notes:validate` | 33 notes valid |

## Docs

No update needed — CLAUDE.md's `epg.events.ts` entry already names `epg-mapping.service.ts` as the home for manual channel-mapping resolution and CRUD, and that is still accurate. Release note added at `.changes/epg-mapping-fail-soft.md` (`type: fix`, area `epg`).

### DIFF
--- a/.changes/epg-mapping-fail-soft.md
+++ b/.changes/epg-mapping-fail-soft.md
@@ -1,0 +1,9 @@
+---
+type: fix
+area: epg
+---
+
+A database error while looking up manual EPG channel mappings no longer breaks
+the request that triggered it. Searching for a channel in the "Map EPG channel"
+dialog, or saving and removing a mapping, now degrades gracefully instead of
+failing outright.
--- a/apps/electron-backend/src/app/events/epg-mapping.service.spec.ts
+++ b/apps/electron-backend/src/app/events/epg-mapping.service.spec.ts
@@ -1,0 +1,246 @@
+const getDatabase = jest.fn();
+const getEpgMapping = jest.fn();
+const getEpgMappingsBatch = jest.fn();
+const setEpgMapping = jest.fn();
+const deleteEpgMapping = jest.fn();
+const searchEpgChannels = jest.fn();
+
+jest.mock('../database/connection', () => ({
+    getDatabase: (...args: unknown[]) => getDatabase(...args),
+}));
+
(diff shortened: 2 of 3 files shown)


4gray:
@codex review

PR #1294, opened 2026-07-27
fix(packaging): stop the Linux frame-copy probe timing out on cold sandboxes

4gray:
## Problem

`Build on linux x64 (flatpak)` fails intermittently at **Verify Flatpak payload, launcher, and sandboxed runtime**:

```
Unable to execute Linux frame-copy runtime probe: spawnSync .../app.asar.unpacked/electron-backend/native/iptvnator_mpv_helper ETIMEDOUT
```

`RUNTIME_PROBE_TIMEOUT_MS: 3000` is a tight budget for a helper that dlopens libmpv plus EGL/GL/GBM, and the Flatpak profile sits closest to that edge because the helper runs inside the sandbox against its bundled closure. On #1277 the job failed three consecutive reruns and passed on the fourth with no code change, while the concurrent master job passed. A successful master run spent 49s on that step versus 37s for a failing one — both walking the line.

## Why not just raise the shared constant

`RUNTIME_PROBE_TIMEOUT_MS` is not CI-only: `embedded-mpv-frame-copy-runtime/probe.ts` uses it for the application's own startup capability decision, a **blocking `spawnSync` on the Electron main process**. Raising it to 15s would let a hung helper stall window creation for 15s on a user's machine — trading a user-facing property for CI stability.

So the two budgets are now separate, because the two callers pay different costs for waiting:

| | Bound | Rationale |
| --- | --- | --- |
| App startup gate | `RUNTIME_PROBE_TIMEOUT_MS` — 3s (unchanged) | Blocks the main process; a timeout degrades gracefully to native-view |
| Packaging verifier | `PACKAGE_VERIFICATION_PROBE_TIMEOUT_MS` — 15s | Nothing waits on it but the CI job's own 120-minute bound; a premature kill reports a healthy package as broken |

## Change

- Two constants join the frozen probe contract: `PACKAGE_VERIFICATION_PROBE_TIMEOUT_MS` (15000) and `PACKAGE_VERIFICATION_PROBE_MAX_ATTEMPTS` (2).
- `runBoundedRuntimeProbe()` retries **only** on `ETIMEDOUT`, repeating the identical bounded launch (same command, args, env, `maxBuffer`, `killSignal`) and announcing the retry on stderr, so a degrading trend stays visible instead of silently absorbed.
- Applied to all profiles, not just Flatpak: a timeout says nothing about the payload regardless of profile, and one code path is easier to trust than two.

## Fail-closed behaviour is unchanged

A hard timeout is the one probe outcome that carries no verdict about the payload. Everything else still fails on the **first** attempt, with no retry:

- spawn error — missing helper, or a wrapper launched instead of the real ELF (`ENOENT`)
- termination by signal (an external `SIGKILL` has no `error`, so it is not mistaken for a timeout)
- nonzero exit
- malformed, multi-line, or wrong-protocol output

A helper that genuinely hangs still fails verification, now after 30s.

## Validation

- `pnpm nx test packaging` → 250/250 pass. `pnpm nx lint packaging` clean.
- `pnpm nx test electron-backend` → 1010/1010 pass (the app-side probe is untouched).
- Four new/extended tests: retry-then-success (asserts the second launch is identical to the first), exhausted timeouts still reject, four non-timeout verdicts each probing exactly once, and the attempt bound itself. **Mutation check:** setting `MAX_ATTEMPTS: 1` fails four of them, so the coverage is not vacuous.
- Verified against real `spawnSync`: a hard timeout sets `error.code === 'ETIMEDOUT'` *and* `signal: 'SIGKILL'`, so the ETIMEDOUT check correctly precedes the signal check.
- End-to-end through the real spawn path against a hanging helper: 2 calls, 30006 ms, one stderr warning, `ETIMEDOUT` returned to the validator.

Docs updated in both canonical contracts (`tools/embedded-mpv/README.md`, `docs/architecture/embedded-mpv-native.md`), including a clarifier at the existing "three-second timeout" line so the two budgets do not read as drift.

No release note: the change touches only `tools/` and `docs/`, which the gate exempts.

## Known residual exposure

The same CI step also runs the **application-level** probe (`flatpak run … --embedded-mpv-runtime-probe`), which still uses the app's 3s budget under the same llvmpipe sandbox. The observed failures were all the verifier probe (the error text and the `app.asar.unpacked/…` path are its), so raising a main-process blocking timeout on speculation seemed the wrong trade. If that second probe starts flaking, the deliberate fix is raising `RUNTIME_PROBE_TIMEOUT_MS` and accepting the startup-stall cost.


4gray:
@codex review

PR #1295, opened 2026-07-27
fix(electron-backend): process zero-delay database cancellation

4gray:
## Summary

- yield one database-worker event-loop turn at each cancellable checkpoint when the artificial batch delay is disabled
- add a real `MessageChannel` regression proving a cancel posted from progress is processed before the operation completes
- document the zero-delay cancellation contract and add a user-facing release note

## Evidence

Before the production change, the regression was stable RED: the worker emitted all progress events, `completed`, and a successful response before the queued cancel task ran. A standalone worker-thread probe showed the same ordering: `progress -> complete -> receipt`.

After the one-line `setImmediate` yield, the queued cancel is received and the operation terminates with `AbortError`. This is a correctness fix, not a claimed performance optimization. It adds no fixed timer delay; the full performance investigation will quantify any cooperative-scheduling overhead.

## Validation

- `pnpm nx test electron-backend --skip-nx-cache --runInBand` — 117 suites / 1173 tests passed
- focused cancellation and database-operation suites — 5 suites / 16 tests passed
- regression test repeated independently — 3/3 passed after stable 3/3 RED before the fix
- `pnpm nx lint electron-backend --skip-nx-cache`
- `pnpm exec tsc -p apps/electron-backend/tsconfig.app.json --noEmit`
- `pnpm nx run electron-backend:build-worker --skip-nx-cache`
- `pnpm run release:notes:validate`
- Prettier and `git diff --check`

### DIFF
--- a/.changes/electron-backend-zero-delay-cancellation.md
+++ b/.changes/electron-backend-zero-delay-cancellation.md
@@ -1,0 +1,8 @@
+---
+type: fix
+area: electron-backend
+---
+
+Cancelling a large Xtream import now stops database work even while it is
+running at full speed, instead of letting the import finish before processing
+the cancel request.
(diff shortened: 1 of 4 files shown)


4gray:
@codex review

PR #1297, opened 2026-07-27
fix(electron-backend): lint the whole project, not one file

4gray:
## The glob

```jsonc
// apps/electron-backend/project.json
"command": "eslint apps/electron-backend/**/*.ts \"...frame-copy-runtime.ts\" \"...frame-copy-runtime/**/*.ts\""
```

The first pattern is unquoted. Nx runs commands through `/bin/sh`, which has no `globstar`, so `**` collapses to `*` and **the shell** expands it — matching **1 file out of 247**. Quoting hands the pattern to eslint, which expands it recursively.

The two sibling patterns were already quoted: the breakage had been worked around per-file instead of fixed. They are now redundant, so all three collapse into one — verified rather than assumed: the quoted glob alone resolves to the byte-identical 247-file list (`diff` empty), including all 24 `frame-copy-runtime` files.

This is the hole that let the 519-line `epg.events.ts` sit over the `max-lines` cap unnoticed.

## What it was hiding

Five errors, now fixed:

**`no-unsafe-finally` × 2 — real bugs.** Both `closeSession` and `disposeSession` did:

```ts
try { await runtime.close?.(); } finally { return this.markClosed(id); }
```

A `return` inside `finally` discards an in-flight exception, so a failing close/dispose was indistinguishable from a clean one. Both *intentionally* always complete teardown — `shutdown()` disposes every session in turn and must not stop at the first failure — so the fix preserves that guarantee with an explicit `catch` and logs the reason instead of dropping it. Same silent-failure class as #1291.

**`prefer-const` × 2.** A write-once `timeoutId` and the module-level `Database`. The *other* `timeoutId` in the same file is genuinely reassigned (`= undefined` on clear) and correctly stays `let`.

**`no-control-regex` × 1.** The recording-filename sanitizer strips control characters on purpose — disabled inline with that rationale rather than reworked.

Three `max-lines` violations would also have surfaced, but #1288 already split those specs, so nothing needed the baseline.

## Tests

Regression coverage for both teardown paths, each **verified to fail against the old code** (reverted the fix, watched the new test go red, restored):

| Spec | Against old code |
|---|---|
| `external-player-session-registry.spec.ts` | 1 failed, 3 passed |
| `embedded-mpv-native.service.spec.ts` | 1 failed, 47 passed |

| Check | Result |
|---|---|
| `pnpm nx lint electron-backend` | **0 errors** (was: 5, hidden), 247 files, 61 pre-existing warnings |
| Full `electron-backend` suite | 105 suites, 1073 tests passed |
| `npx prettier --check` on every changed file | clean |
| `pnpm run release:notes:validate` | 36 notes valid |

## Note on the warnings

61 warnings remain and are left alone: they are pre-existing, `--max-warnings` is not set anywhere, so they do not gate CI. Cleaning them is separate work.

Release note is `type: internal` — teardown finishes exactly as before, only the previously-discarded reason now gets logged.

### DIFF
--- a/.changes/electron-backend-lint-coverage.md
+++ b/.changes/electron-backend-lint-coverage.md
@@ -1,0 +1,9 @@
+---
+type: internal
+area: electron-backend
+---
+
+Lint now covers the whole Electron backend instead of a single file, and two
+teardown paths that silently discarded their failure reason — closing an
+external player session and disposing an embedded mpv session — now report it.
+Both still finish the teardown exactly as before.
--- a/apps/electron-backend/project.json
+++ b/apps/electron-backend/project.json
@@ -241,7 +241,7 @@
                 "{workspaceRoot}/tools/embedded-mpv/runtime-probe-contract.cjs",
                 "{workspaceRoot}/tools/embedded-mpv/runtime-probe-contract.d.cts"
             ],
-            "command": "eslint apps/electron-backend/**/*.ts \"apps/electron-backend/src/app/services/embedded-mpv-frame-copy-runtime.ts\" \"apps/electron-backend/src/app/services/embedded-mpv-frame-copy-runtime/**/*.ts\""
+            "command": "eslint \"apps/electron-backend/**/*.ts\""
         },
         "test": {
             "executor": "@nx/jest:jest",
--- a/apps/electron-backend/src/app/events/epg-worker.service.ts
+++ b/apps/electron-backend/src/app/events/epg-worker.service.ts
@@ -422,7 +422,6 @@ export class EpgWorkerService {
             }
 
             let settled = false;
-            let timeoutId: ReturnType<typeof setTimeout>;
             const settle = (fn: () => void) => {
                 if (settled) return;
                 settled = true;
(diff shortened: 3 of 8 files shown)


4gray:
@codex review

4gray:
Superseded by #1176, which landed while this was open and fixed the same things independently: it quoted the lint glob and repaired every violation that exposed. `nx lint electron-backend` on master is now 0 errors across 249 files, so nothing here is still needed.

Closing rather than rebasing — the overlap is near-total, and the two remaining differences are not worth a PR of their own on their honest merits:

- **Logging the discarded reason.** #1176 used a bare `catch {}` in `closeSession` and `disposeSession`; this PR logged the failure. The `no-unsafe-finally` bug is fixed either way, but the reason a close/dispose failed is still dropped silently on master. Worth a small follow-up, not a rebase of this branch.
- **Collapsing the three lint patterns into one.** Verified redundant here (the quoted glob alone resolves to the identical 247-file set), but purely cosmetic.

For the record, since it may be useful if that code is revisited: both teardown paths intentionally always complete, because `shutdown()` disposes every session in turn and must not stop at the first failure. Any future change there needs to preserve that.

PR #1298, opened 2026-07-27
refactor(portals): share external-button and position-writer logic

4gray:
A behaviour-preserving refactor that removes real duplication between the Xtream
and Stalker VOD detail views, and brings three files back under the 400-line
ESLint limit — none of which were baselined, so they had no headroom left.

## What moved

**1. `createExternalPlaybackButtonState` → `libs/portal/shared/util`**

Xtream and Stalker each carried a near-identical private copy of six computeds
deriving the Play/Stop button state from the active external (MPV/VLC) session.
A Play button that behaves differently per portal is a bug waiting to happen, so
both now read from one implementation.

**2. `createInlinePlaybackPositionWriter` → same lib**

Both portals had their own copy of the "persist the position, but at most once
per 15s" logic. Same reasoning: a resume point that drifts per portal is exactly
the kind of divergence users notice.

**3. `startStalkerVodDownload` → `stalker-vod-download.ts`**

Lifted out of the Stalker detail component as a self-contained errand: resolve a
playable link (Stalker hands out `create_link` URLs, not static ones) and hand
the result to the download manager. Nothing else in the component needed it.

## Effect on the three files at the hard limit

| File | Before | After |
| --- | --- | --- |
| `libs/ui/playback/.../vod-details/vod-details.component.ts` | 389 | 333 |
| `libs/portal/stalker/feature/.../stalker-catalog-detail.component.ts` | 394 | 325 |
| `libs/portal/xtream/feature/.../vod-details-playback.service.ts` | 345 | 275 |

Net −247 lines of product code, +33.

## Tests

16 new tests for the two shared helpers. The one that matters most asserts that
a session playing a *different* movie must not turn this page's Play button into
Stop — the failure mode the shared helper exists to prevent. The rest cover
playlist/content-type mismatch, closed and errored sessions, the launching
state, throttle behaviour, and `reset()`.

## Behaviour

Deliberately neutral — the helpers are equivalent to the copies they replace.
Two cosmetic simplifications worth a reviewer's eye:

- The Xtream copy had an extra `if (!playlistId) return null` guard that the
  shared helper drops. It is unreachable as a behaviour difference:
  `PlayerContentInfo.playlistId` is a required `string`, so when the store's
  playlist id is undefined the `!==` comparison already yields `null`.
- `startStalkerVodDownload` normalizes `item.cmd` to `''` instead of passing
  `undefined` through to `fetchLinkToPlay`, whose parameter is typed `string`.

## Known follow-up

The `resumeSettled` latch that exists on the Xtream side in #1286 is **not** in
the shared writer here — pulling it in would have smuggled a behaviour change
into a refactor. Once both land, it should move into the shared writer and
benefit Stalker too.

This refactor is the prerequisite for the Stalker multi-source phase.

## Validation

Rebased cleanly onto `origin/master` (`bfad82c2`); no file overlap with anything
master touched since the branch point, and the patch is byte-identical
post-rebase.

```
pnpm nx run-many --target=lint,test --projects=portal-shared-util,portal-stalker-feature,portal-xtream-feature,ui-playback
```

→ all 8 targets green. `portal-shared-util` 87 tests / 13 suites (16 new),
`ui-playback` 725 tests / 83 suites.

```
pnpm nx build web
```

→ succeeded (pre-existing SCSS budget and shaka-player CommonJS warnings only).

No E2E run: no user-visible workflow, route, persistence or IPC behaviour
changes. No `.changes/` note — pure refactor, so the PR carries
`no-release-note` per `.changes/README.md`. No doc updates needed; the extracted
helpers are internal and `CLAUDE.md`'s descriptions of these areas remain
accurate.

### DIFF
--- a/libs/portal/shared/util/src/index.ts
+++ b/libs/portal/shared/util/src/index.ts
@@ -26,3 +26,5 @@ export * from './lib/navigation/portal-rail-links';
 export * from './lib/navigation/portal-route.utils';
 export * from './lib/navigation/workspace-portal-navigation';
 export * from './lib/collection';
+export * from './lib/external-playback-button-state';
+export * from './lib/inline-playback-position-writer';
--- a/libs/portal/shared/util/src/lib/external-playback-button-state.spec.ts
+++ b/libs/portal/shared/util/src/lib/external-playback-button-state.spec.ts
@@ -1,0 +1,138 @@
+import { signal } from '@angular/core';
+import type { ExternalPlayerSession } from '@iptvnator/shared/interfaces';
+import { createExternalPlaybackButtonState } from './external-playback-button-state';
+
+function session(
+    overrides: Partial<ExternalPlayerSession> = {}
+): ExternalPlayerSession {
+    return {
+        player: 'mpv',
+        status: 'playing',
+        contentInfo: {
+            playlistId: 'playlist-1',
+            contentXtreamId: 42,
+            contentType: 'vod',
+        },
+        ...overrides,
+    } as ExternalPlayerSession;
+}
+
+function setup(initial: ExternalPlayerSession | null = null) {
+    const current = signal<ExternalPlayerSession | null>(initial);
+    const api = createExternalPlaybackButtonState({
+        session: current,
+        playlistId: signal<string | null>('playlist-1'),
+        contentId: signal<number | null>(42),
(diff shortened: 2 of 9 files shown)


4gray:
@codex review

PR #1305, opened 2026-07-28
test(performance): preserve delayed worker ELD samples

4gray:
## Summary

- require two event-loop-delay histogram polling turns before applying the 50 ms arming deadline, while retaining the unconditional 50-poll ceiling
- leave request work, response posting, and histogram flush behavior unchanged
- add deterministic regressions for a delayed first turn both with and without a second-turn sample, and document the arming/flush contract

## Evidence

A formal synthetic Xtream baseline capture exposed one non-overlapping `DB_SET_APP_STATE` request whose first arming timer ran 86.202 ms late. `monitorEventLoopDelay` seeds on its first native turn and exposes the first sample only after a second turn, so the old elapsed-time check reported `event-loop-delay-arm-timeout` before that required second poll. A scan of the preserved local evidence found this exact condition once across 5,161 request records; production database behavior was unaffected.

The change is limited to development/test-only opt-in profiling and its tests/docs. Raw traces, CPU profiles, and heap snapshots remain under git-ignored `dist/performance/` and are not committed.

## Validation

- `pnpm nx test electron-backend --skip-nx-cache --runInBand --testPathPatterns='worker-performance-capture'` — 4 suites, 18 tests passed
- `pnpm nx test electron-backend --skip-nx-cache --runInBand` — 125 suites, 1,261 tests passed
- `pnpm nx lint electron-backend --skip-nx-cache` — passed with 0 errors (existing warnings only)
- Prettier and `git diff --check` — passed
- synthetic Xtream smoke on the local mock server — 5 scenarios, 15 warm-up/measured/diagnostic captures; all contracts valid, no ELD arming timeouts, no profiling failures, no BrowserWindow unresponsive events

## Release note

Not required: this changes only the opt-in development/test performance profiler and its documentation.

### DIFF
--- a/apps/electron-backend/src/app/workers/worker-performance-capture.concurrency.spec.ts
+++ b/apps/electron-backend/src/app/workers/worker-performance-capture.concurrency.spec.ts
@@ -6,42 +6,13 @@ import {
     releaseDatabaseWorkerPerformanceCapture,
     startWorkerPerformanceCapture,
     type WorkerPerformanceCapture,
-    type WorkerPerformanceCaptureRuntime,
 } from './worker-performance-capture';
-import { DEFAULT_WORKER_PERFORMANCE_RUNTIME } from './worker-performance-capture.runtime';
 import { createFakeRuntime } from './worker-performance-capture.test-harness';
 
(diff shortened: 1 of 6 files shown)


4gray:
@codex review

4gray:
@greptileai

Issue #1118, opened 2026-07-02
Live TV layout: channels + EPG column gets squeezed while the player panel dominates


PavluntiyJ:
## Context

On the Live TV view (Xtream/Stalker) the screen is three columns: Groups (shell context panel), Channels (live-layout sidebar with EPG previews), and the Details/Player area. Channels + EPG is where users spend most of their browsing time, but it's currently the most constrained column, while the player area — secondary for users on external players (MPV/VLC) — always takes the remaining majority of the width.

## Specific problems

1. **EPG disappears instead of degrading.** `channel-list-item` has container-query breakpoints that hide the EPG title, progress bar and program-info button entirely below 270px, and the logo/actions below 220px. The sidebar's `minWidth` is 250px — right at the cliff edge, so a slightly-narrow sidebar silently loses all EPG info. Adaptive shrinking (drop end-time first, then compact type) would preserve the most valuable info at any width.

2. **The player panel is not collapsible as a whole.** The channels sidebar can collapse, the inner EPG timeline can collapse, but a user who plays everything in MPV/VLC cannot give the freed space to the channels column.

3. **Groups cannot collapse independently.** The shell context panel only collapses together with the channels sidebar via the single "everything quiets" toggle (`LiveLayoutSidebarStateService`). There is no way to hide just the Groups rail, and it has a generous max width (560px) for what is secondary navigation.

4. **Default widths favor the player.** Channels defaults to 400px / max 600px regardless of viewport; on wide screens the player area gets everything else.

## Proposal (open to discussion)

- Channels: default 460px, min 340px (safely above the EPG degradation thresholds), max `min(760px, 45vw)`.
- Groups: default 260px, min 220px, max 340px, plus an independent collapse toggle (same chevron/restore pattern the channels sidebar already has).
- Details/Player: collapsible as a whole — collapse via width/visibility, not `display:none`, so an embedded video element keeps playing and restores without re-initializing; the freed width goes to the channels column. One caveat found while prototyping: a floating restore affordance at the top-right of the content area ends up overlapping the sidebar-header buttons once the sidebar takes the full width, so the toggle is better placed in the sidebar header itself.
- Below ~1024px viewport width, Groups collapses by default.
- Keep the existing `appResizable` directive; add a thin coordination layer at the shell level so simultaneous panels respect each other's min widths instead of relying only on static per-panel maxWidth.

[...]

Issue #1128, opened 2026-07-04
[FEATURE] Add a "Catch-up available" indicator for live channels that have them (XC)


WolfganP:
**Is your feature request related to a problem? Please describe.**
Now that a lot of work was added to managing the EPG and schedule (thx a lot, btw), it would be interesting to have an indicator in the channels list to mark the ones catched up channels by the service provider.

**Describe the solution you'd like**
For example, a symbol besides the current program info on the line below for the channels with EPG, below the star on the right side

Thanks!

PR #1312, opened 2026-07-30
fix(ui): preserve EPG in narrow channel rows

4gray:
## What changed

- Keep the programme title, progress, and enabled channel actions in narrow EPG rows.
- Apply the agreed priority cascade: hide the end time at `310px`, the logo at `270px`, and the start time at `220px`.
- Align loaded rows, skeletons, and fixed virtual-scroll strides at `68px` with EPG and `52px` without EPG across M3U, Xtream, Stalker, Favorites, and Recent views.
- Keep radio/PWA rows compact without false EPG placeholders, while preserving the fixed stride of mixed M3U lists.
- Hide the inapplicable `Map EPG` context action in Stalker Radio.

## Why

The previous narrow layout dropped useful programme context and actions exactly when the channel panel became constrained. This is the first, low-risk increment from the broader workspace-layout proposal: it fixes information loss without coupling it to the larger independent-panel behavior.

Related to #1118. This intentionally does not close the issue; independent Groups/Channels toggles and reclaiming the external-player/right-side empty region remain follow-up work.

## Release note

- [x] Added `.changes/ui-narrow-channel-epg.md`
- [ ] Not needed

## Checks

- [x] Unit/regression coverage added or updated in the affected Components, M3U, Xtream, Stalker, portal shared UI, portal shared util, workspace shell, and mock-server projects
- [x] Affected project lint targets pass (only pre-existing warnings remain)
- [x] `pnpm nx build web --skip-nx-cache`
- [x] Electron E2E: `xtream-epg.e2e.ts` — 5/5, including real `300/232/200px` geometry and Stalker Radio; focused post-review Radio rerun — 1/1
- [x] Manual Electron/CDP verification with `agent-browser` in light and dark themes; no renderer errors
- [x] `pnpm run release:notes:validate` and `git diff --check`


WolfganP:
Thx @4gray 
What a shame I couldn't catch your work on this topic earlier, my request on catch up indicator https://github.com/4gray/iptvnator/issues/1128 would have been a good addition to this PR
Thx again for all the improvements

PR #1321, opened 2026-08-01
fix(playback): escape commas in mpv http header fields

### LINKED ISSUE #910: Stalker Live TV dont work except in VLC
Good stalker working on Tivimate, Stalker IPTV or whatever app. All player (Artplayer, Video.js. HTML5) dont work. 

Tried .ts and m3u format. 

Only way to get it working is by puttinh external player VLC. 

Using Windows X64 version 0.21.

wretchwatson:
Fixes #910

## Problem

Stalker live TV fails to play in the embedded mpv player (only works in VLC / external players).

`mpv --http-header-fields` is parsed as a **comma-separated list** (`OPT_STRINGLIST`). Stalker portals require the MAG250 user agent, which contains a comma:

```
Mozilla/5.0 (QtEmbedded; U; Linux; C) AppleWebKit/533.3 (KHTML, like Gecko) MAG250
```

Because the value is split on commas, the `X-User-Agent` header was truncated to `... (KHTML`. The portal then rejects the stream request with `HTTP 400 Bad Request` — while VOD (`movie.php`) still works because it does not validate this header.

## Fix

Escape backslashes and commas inside header values (`\,`). mpv's stringlist parser strips the backslash and keeps the comma as part of the header value, so the full MAG250 user agent is sent to the portal.

## Verification

- `curl` against the live portal: intact `X-User-Agent` → `HTTP 302` (accepted); truncated header → `HTTP 400`/`462` (rejected)
- After the fix, the spawned mpv process receives the escaped header and the live channel plays

### DIFF
--- a/.changes/stalker-mpv-header-comma-escape.md
+++ b/.changes/stalker-mpv-header-comma-escape.md
@@ -3,9 +3,8 @@ type: fix
 area: playback
 ---
 
-Stalker live TV channels play again in the embedded MPV player. The MAG250
-user agent sent to portals contains a comma (`... (KHTML, like Gecko) MAG250`),
-and MPV splits its `--http-header-fields` list on commas, so the `X-User-Agent`
-header was truncated and strict portals rejected the stream with HTTP 400.
-Commas and backslashes in header values are now escaped before MPV parses them,
-on the Windows/Linux and macOS embedded-MPV paths.
+Stalker live TV plays again in the embedded MPV player. The MAG250 user agent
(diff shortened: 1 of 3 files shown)


wretchwatson:
Addressed the review: the macOS embedded-MPV implementation in embedded_mpv.mm has its own duplicate joinHeaderFields, now updated with the same backslash/comma escaping (commit eabda32).

4gray:
@greptileai review

PR #1326, opened 2026-08-01
fix(ui): make the workspace usable on phone-sized screens

### LINKED ISSUE #1100: UI broken on mobile device
**Describe the bug**
I am using the latest docker version and UI looks very broken. Menu is located out of the screen

**To Reproduce**

1. Just open the app on a smaller screen like mobile device.

**Expected behavior**
Menu should be located on the left side and all should be more responsive to lover resolutions.

**Screenshots**
Image

-   Docker container
-   OS: Linux
-   Browser: chrome
-   Version: 0.22.0
-   Feature: Xtream Code

**Additional context**
No additiona context

4gray:
Closes #1100.

The shell was already half-adapted below 640px — the grid restacks, the rail flips to a horizontal bar — but several pieces never got the memo, and the result was the broken screenshot in the issue.

## What was actually wrong

**Navigation drawn outside its own bar.** `.app-rail` turns into a horizontal top bar at ≤640px, but the nested `app-workspace-shell-rail-links` kept `flex-direction: column` with no breakpoint of its own, so its links stacked downwards out of the 52px bar and over the header. `.rail-context-region` also kept its vertical `max-width: 48px`, which piled every portal section on top of itself.

**Three rails wider than the screen.** The shell context panel (300px), the live-layout channel sidebar (400px) and the M3U channel drawer (460px) all get their width from `ResizableDirective`, which writes the persisted desktop width as an inline style. On a 375px screen that left the content around 50px. They now span the full width and stack above the content — and because the width is inline, the rules need `!important`, with `.resize-handle` hidden in the same breath since dragging is meaningless at full width.

**Header overflow.** 458px of content in a 375px row. The keyboard-only affordances (the `⌘K` badge, the shortcuts button) are dropped on a screen that usually has no keyboard; the command palette still opens via the real shortcut and the rail.

## Found while walking the rest of the UI

Checked dashboard, sources, settings (all seven sections), VOD/series/live, detail pages, search, collections, dialogs and inline playback at 375px, 768px and desktop:

- **The `Play` button on a movie page collapsed to its icon.** The hero kept poster and details side by side, leaving the info column ~199px, and the action row shrank below its own labels.
- **The settings section list painted over its footer.** `height: 100%` does not resolve against a stretched flex item, so the list kept its full content height. This also affected short desktop windows, not just phones.
- **Hiding the M3U channel list on a phone was one-way.** The mobile rule hid the restore handle while the collapse toggle stayed reachable by touch, leaving `Cmd/Ctrl+B` as the only way back.
- **The live header drew the channel count and the paginator on top of each other** — visible up to tablet width. The paginator does not shrink, so `.category-meta` collapsed to zero width and its nowrap text overflowed the box. Fixed by letting the paginator wrap and shrink (its own container already wraps internally) and by clipping the meta.
- **The search scope checkbox was pushed off the right edge.**

## Live TV

Two lists stack above the player there. Capping both still left the video a sliver, so the player container states a `min-height` and the lists shrink into what is left. The existing collapse toggle then gives a genuine full-screen viewing mode on a phone.

## Validation

Lint clean (0 errors), unit tests for every touched project plus `web` (197 tests) green, and the production PWA build compiles the SCSS. Desktop and tablet were re-checked for regressions: the vertical rail, side panels and single-row paginator are unchanged.

Verified in-browser at 375×812 across every route above, including inline VOD playback and live playback against the Xtream mock server.

## Known limitation, deliberately out of scope

The categories panel still occupies 30vh on phones even on detail pages and during playback, where it earns nothing. The right fix is an off-canvas drawer with a header toggle — new control, TS and i18n — which is a larger change than this issue needs. Nothing is unreachable today; it just costs space.

Separately: this issue is about layout and layout is fixed, but *using* the PWA against real providers still runs into CORS/mixed-content, which is a property of PWA mode rather than something CSS can address.

### DIFF
--- a/.changes/ui-mobile-layout.md
+++ b/.changes/ui-mobile-layout.md
@@ -1,0 +1,10 @@
+---
+type: fix
+area: ui
+issues: [1100]
+---
+
+IPTVnator is usable on a phone again. The navigation bar no longer sits off
+screen, category and channel lists take the full width instead of squeezing
+the content into a sliver, the video keeps a usable share of the screen, and
+the M3U channel list can be reopened after hiding it.
--- a/apps/web/src/app/settings/settings.component.scss
+++ b/apps/web/src/app/settings/settings.component.scss
@@ -751,6 +751,9 @@
     );
     margin-left: 0;
     margin-right: 0;
+    // Stretched to the full layout width, so its own inline padding would be
+    // added on top and push the settings page into a horizontal scroll.
+    box-sizing: border-box;
     z-index: 100;
     display: flex;
     justify-content: flex-end;
--- a/docs/architecture/iptvnator-ui-guidelines.md
+++ b/docs/architecture/iptvnator-ui-guidelines.md
@@ -253,9 +253,11 @@ remain local when the meaning is explicit.
   width is preserved so uncollapsing restores the user's previous resized
   width. Both rails share the same 180 ms width transition so motion stays in
   lockstep.
-- Below 600 px viewport, the M3U layout's mobile bottom-drawer rule overrides
-  the desktop collapse to `height: 0` instead of `width: 0`, and the floating
-  restore handle is hidden.
+- At the phone breakpoint the M3U layout's bottom-drawer rule overrides the
+  desktop collapse to `height: 0` instead of `width: 0`. The floating restore
+  handle stays visible there: the collapse toggle is reachable by touch, so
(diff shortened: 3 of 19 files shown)


4gray:
@codex review

4gray:
@codex review

4gray:
@codex review

4gray:
@codex review

4gray:
@codex review

Issue #1161, opened 2026-07-10
Can't search for "A&E" in global search


kurtdh:
If you go to global search and search for "A&E" without the quotes of course, the channel will not show up despite it being available. I think it has something to do with the & character being in the search.

stunner999:
I can also confirm this live channels using Stalker Portals Global search doesn't work because live channel are not even loaded.

PR #1336, opened 2026-08-01
feat(search): chip input and chip-based matching for advanced search

ilyesbrh:
## What changed

The global **Advanced search** bar becomes a chip input, and matching is now
chip-based:

- Type a phrase and press **Enter** to commit it as a visible, removable chip.
  Short chips like `FR` are allowed; draft text does not search until committed.
- Each chip is a **joined** unit — all of its words must be present in a title.
  A result matches if it satisfies **any** chip, ranked so results matching the
  most chips come first (single-chip matches still appear).
- A plain one-chip query behaves exactly as before, so existing `?q=` links and
  the tuned matcher (compound words `#1161`, diacritics) are unaffected.

Scope is intentionally the global Advanced search bar only — portal and playlist
search bars keep their plain text input. Chips serialize as newline-delimited
units over the existing string query, so the facade, `?q=` param and IPC
contract stay string-typed (no new bridge methods).

Known limitation (called out for reviewers): within a chip, a short word (≤2
chars) anchors to the title start via the existing matcher — e.g. `[fr bein]`
matches `FR beIN 1` but not `BeIN Sports FR`. Broadening short words to match
mid-title would require reworking candidate SQL and is left as a follow-up.

## Why

Users want to search with several independent terms (e.g. `FR` + `beIN` +
`1968`) and see the closest matches first, instead of one query forcing every
word to appear in a single title.

## Release note

- [x] Added a note under `.changes/` (`search-token-partial-match.md`)
- [ ] Not needed

## Checks

  - `content-search.util.spec.ts` (chip parsing + chip scoring)
  - `content.operations.spec.ts` (multi-chip OR ranking end-to-end)
  - `workspace-shell-header.component.spec.ts` (chip add/remove/dedupe/render)
- [x] `pnpm run lint` (electron-backend, workspace-shell-feature) and the
  affected `pnpm nx test` targets pass


ilyesbrh:
Thanks @greptile-apps — the M3U ranking issue is valid and fixed.

`scoreM3uChannel` no longer scores each field independently and `Math.min`s them. `scoreGlobalSearchChips` now accepts the searchable fields (name, TVG name, group title) as one set and counts a chip as matched when it hits **any** field, so a channel satisfying different chips in different fields is ranked by the total distinct chips matched. Xtream (single title) behaviour is unchanged.

Added a regression test in `content-search.util.spec.ts` ("counts a chip as matched when it hits any of several fields"). Folded into the `chip-based OR matching` commit.

PR #1358, opened 2026-08-02
fix(stalker): anchor auth-failure body detection and share it across transports

4gray:
Split out of PR 6 (#1354) so the auth-failure detector lands on its own — it is the one piece both #1344 and #1354 need, and unifying it first keeps #1354 focused and shrinks its rebase.

## The bug

The stock middleware answers auth failures with HTTP 200 and a **bare plain-text body** (`Authorization failed.`, `Access denied.`, `Unauthorized request.`). `isStalkerAuthFailureBody` matched those by **substring** under a 200-character cap, so anything short enough that merely *contained* one of the phrases was read as the portal refusing authorization:

| Body | before | after |
| --- | --- | --- |
| `Authorization failed.` | ✅ | ✅ |
| `Authorization failed. 75` (stock counter) | ✅ | ✅ |
| `Access denied.` / `Unauthorized request.` | ✅ | ✅ |
| `<html><body>Access denied</body></html>` (38 chars) | ✅ | ❌ |
| `Error: access denied by policy` | ✅ | ❌ |
| `The request Authorization failed. Try again later.` | ✅ | ❌ |

The length cap only stops *long* HTML pages. A short one from something in **front** of the portal — a reverse proxy, a WAF, a captive portal — sails through, and this predicate feeds two decisions from #1344: `classifyStalkerProbeResponse` (is this endpoint token-enforcing?) and `shouldAttemptRepair` (is the persisted config wrong?). So a portal that never answered at all could be re-probed and reclassified because its proxy said "Access denied".

The match is now anchored to the whole trimmed reply, keeping the stock server's optional trailing counter. The structured `js.error`/`js.msg` fields keep the **wider** phrase set (`Invalid token`, `Auth failed`, bare `unauthorized`) — a panel fills those in deliberately, so breadth is safe there; that asymmetry was already the intent and is now enforced by the shapes rather than by a length cap.

## The move

Detection goes to `@iptvnator/shared/interfaces` (`stalker-auth-failure.util.ts`). It has to live somewhere both transports reach: the **Electron main process is where these bodies actually arrive**, and it cannot import a renderer library — the same reason #1348 centralised the identity and URL builders. `stalker-portal-discovery.utils.ts` re-exports the three functions, so no call site changes and the existing specs pass untouched.

One shape added while consolidating: `{js: '<phrase>'}` (phrase directly in `js`, not in an object). `StalkerSessionService.isAuthorizationError` already accepted it via its `JSON.stringify` pass, so this preserves behavior that the structured detector would otherwise have dropped.

## Tests

New `stalker-auth-failure.util.spec.ts` — the documented bodies, the case/whitespace/counter variants, the three false positives above, the JSON envelope forms, `{js: '<phrase>'}`, and the message helper. The existing `stalker-portal-discovery.utils.spec.ts` cases are unchanged and still pass, which is the compatibility check that matters.

| Command | Result |
| --- | --- |
| `nx test shared-interfaces` | 194 passed |
| `nx test portal-stalker-data-access` | 278 passed |
| `nx run-many -t test -p portal-shared-data-access,services,playlist-import-feature,electron-backend` | 1994 passed |
| `nx run-many -t lint -p shared-interfaces,portal-stalker-data-access` | 0 errors |

Docs updated where the wire format is described. Follow-up in #1354: the Electron transport starts *producing* a structured marker for these bodies, which is added to this same shared function rather than a second copy.

### DIFF
--- a/.changes/stalker-auth-failure-detection.md
+++ b/.changes/stalker-auth-failure-detection.md
@@ -1,0 +1,8 @@
+---
+type: fix
+area: stalker
+---
+
+A Stalker portal is no longer re-probed and reclassified because something
+in front of it — a proxy or firewall, not the portal — answered with a short
+"Access denied" page. Only the portal's own authorization replies count now.
--- a/docs/architecture/stalker-portal.md
+++ b/docs/architecture/stalker-portal.md
@@ -71,7 +71,17 @@ Two portal modes exist, persisted per playlist as
   except `handshake`, `get_profile`, `get_localization`, and `do_auth`
   requires `Authorization: Bearer <token>`; auth failures are HTTP 200 with a
   plain-text body (`Authorization failed.`, `Access denied.`,
-  `Unauthorized request.`), never a 401/403. While a full portal is the
+  `Unauthorized request.`), never a 401/403. Detection of those bodies — and
+  of the JSON envelope (`{js: {error|msg}}`) some panels answer instead —
+  lives in `libs/shared/interfaces/src/lib/stalker-auth-failure.util.ts`, so
+  the Electron main process (where the bodies actually arrive) and the
+  renderer classify identically; `stalker-portal-discovery.utils.ts`
+  re-exports it. The body match is anchored to the whole reply: these are
+  bare phrases, and a substring rule also accepted a short proxy or WAF page
+  containing one, which then triggered portal reclassification against a
+  portal that never answered. The structured `js.error`/`js.msg` fields keep
+  the wider phrase set (`Invalid token`, `Auth failed`, bare `unauthorized`),
+  since a panel fills those in deliberately. While a full portal is the
   active playlist, `StalkerSessionService` keeps a **watchdog** running —
   periodic authenticated `watchdog/get_events` pings (currently every 25 s;
   the protocol default expects 120 s, tracked for a later PR) whose failures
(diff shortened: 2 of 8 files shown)


4gray:
@codex review

4gray:
Good catch, and it was a real hole in this PR's own claim — fixed in 59d4060af.

Verified the chain before changing anything, because it runs through three layers:

1. A short page from something in front of the portal (`The request Authorization failed. Try again later.`) correctly fails the newly anchored body detector…
2. …but `isAuthorizationError()` then stringified the whole response and matched an **unanchored** `authorization failed`, so the session retired the token and retried;
3. the retry threw `Error('Authorization failed after retry')`, and *that message* matches the wide phrase set in `shouldAttemptRepair` — so the repair layer re-probed and could reclassify a portal that had refused nothing.

Exactly the false positive this PR exists to remove, reached through a different door. The sweep is gone; the shared detector already covers every real shape including the `js.error`/`js.msg` envelopes it was also catching, so no coverage is lost. What remains in that method is structural only: the shared detector, the controlled-message check, and the HTTP 401/403 codes.

Per your note the regression goes through `makeAuthenticatedRequest`, not the primitive — that is where the chain actually ran. It asserts the response passes through untouched, the token stays cached, and **no retry is issued** (so nothing reaches the repair trigger). I verified it fails against the pre-fix code before keeping it.

Validation: 928 unit tests across 4 affected projects, lint 0 errors, `nx build web` clean.

Note for whoever merges this alongside #1354: that PR replaces this method entirely with `stalker-response-classification.ts`, which already has no stringify sweep — so the two agree, and the conflict there is a straight deletion.

@codex review
@greptileai

PR #1359, opened 2026-08-02
feat(portals): rework the VOD sources popup and action row

4gray:
## Summary

Implements the `design_handoff_vod_sources_popup` handoff for the Xtream VOD detail view — recreated with our Angular components and theme tokens, not by porting the prototype's HTML.

Two problems drove it: the sources popup could clip off-screen in short windows, and the action row's four equal-width labeled buttons made every secondary compete with Play.

## Changes

**Popup anchoring & fit** — opens above the chip (right edges aligned, with a pointer tail). CDK flexible dimensions cap it to the space beside the anchor, so only the source list scrolls and the header, search, filter chips and footer stay visible at any window size; it flips below the chip when less than the overlay `minHeight` remains above, and re-applies on resize. Closes on outside click and Escape.

**Filter chips** — All (N) / Available / HD+ / language, composing AND with the playlist search, with an "X of N" counter when the list is reduced. Two deliberate calls:
- "Available" is strict (`probe.status === 'ok'` only) — an unchecked source must never pass a filter with that name. Since checks are lazy, activating it while nothing anywhere has a verdict runs check-all itself.
- "HD+" accepts parsed quality as well as stated. A narrow, documented exception to "guesses never drive decisions": it's a browse aid operated while the `~` markers are visible, and hiding a copy whose own title says 1080p would look broken. Ranking and failover still go through `factualOnly()`.

**Copy rows** — expanding "N copies in this playlist" no longer repeats the playlist domain or monogram. Each copy shows a parsed language chip (`EN|`, `RU|`…), the provider's raw stream title in mono, and only the tags that *differ* from the parent copy; the parent's own copy reads "same as above". Tag rendering moved into a shared `VodSourceTagListComponent` so the full row and the copy row can't drift on how they present a guess vs a fact.

**Bounded, remembered checks** — availability checks run at most 4 concurrently (`createCheckQueue`); a row flips to "checking" the moment it's asked for, which also dedupes a second click while it waits for a slot. Settled verdicts (`ok`/`fail`, never `unknown`) are cached 10 minutes per movie+source in the root `VodSourceProbeCacheService` and seeded back after discovery — reopening a movie shows its verdicts without contacting any portal. In-memory only: a probe result is a claim about the present.

**Action row** — Favorites and Download become icon-only 64px buttons at Play's height: filled red heart with a scale pulse when favorited; download idle icon → SVG progress ring (real percent, indeterminate spin when the size is unknown, resume glyph when paused) → green checkmark that reveals the file. State comes from the download manager, so it survives navigation. Clicking a running download asks before cancelling.

**Chip badge** now counts total copies across playlists (what the popup lists, matching the "All (N)" chip); the in-player chip still counts alternatives, since its sentence is "N other places to go".

## Behavior change worth flagging

The labeled "Play from source" secondary shown for downloaded movies is **removed** per the design — the done-state checkmark takes that slot, and provider playback for a downloaded movie goes through the Sources popup. Specs updated accordingly.

## Testing

- [x] Unit: 596 tests green across `components`, `portal-xtream-feature`, `portal-shared-data-access` — including new suites for the filter predicates, the copy-row diffing, the concurrency queue, and the probe cache's TTL/"never cache unknown" rules. `ui-playback` (915) green as a consumer check.
- [x] E2E: `electron-backend-e2e:e2e-ci--src/vod-multi-source.e2e.ts` green on a fresh build (badge assertion updated for the new total-copies count).
- [x] Verified in a real Electron run — the light-theme pass caught white-alpha button surfaces that vanish over the bright hero, fixed by moving the icon buttons and the chip onto the theme-aware `--tag-*` tokens.
- [x] i18n: 10 new keys across `en.json` + all 18 locales via `tools/i18n/fill-missing.mjs`; coverage, placeholder-integrity and `nx build web` gates pass.
- [x] Lint green; release note added and validates.
- [x] Docs: `docs/architecture/vod-multi-source.md` gained sections on anchoring/fit, filters, copy rows and bounded checks; the `CLAUDE.md` multi-source bullet synced.

Not covered locally: a dark-theme visual pass (same long-established `--app-*`/`--tag-*` families the rest of the panel uses), and a screenshot of expanded copy rows — the mock catalogs carry one copy per playlist, so that path is covered by unit tests instead.


4gray:
Thanks — one finding was real and is fixed; the other does not hold. Details so the reasoning is on the record.

**P2 "Preserve provider playback for single-source downloads" — valid, fixed in ea505a6a9.**

Confirmed the gap: once `isDownloaded()`, the primary button routes to `playFromLocal()`, and the sources chip is behind `@if (multiSource.hasAlternatives())`. With a single playlist there was consequently no path left to stream the provider copy. Restored as an icon button beside the downloaded checkmark, gated on `externalPrimaryButtonState() === 'idle'` exactly as the previous labeled action was, plus a regression test (`still offers provider playback for a downloaded movie with no alternatives`).

**P2 "Capture the probe session before enqueueing checks" — not reproducible; deliberately left as is.**

The described state cannot occur: in `discover()` the `sessionToken++` and `this.controller = new VodMultiSourceController()` sit in the same synchronous `if (!sameMovie)` block with no `await` between them, so an old controller can never pair with a new token. Read at dequeue time both are always a matching pair, and a task belonging to an abandoned movie asks the *current* controller for its source id — `findSource` returns null (ids embed the provider content id, so they do not collide across films) and `probeSource` returns before any resolve or HEAD request.

I implemented the suggestion anyway to check, and it actively introduces a bug. `sessionToken` also advances when the movie identity is momentarily null, and **that path does not replace the controller** — only a different film does. A task dropped on a stale captured token therefore abandons a row it had already marked `probing` in the controller still on screen, leaving the spinner stuck forever. Reading late instead completes the check, which is the correct outcome when the user returns to the same movie.

Evidence: a new test that queues past the concurrency limit and then opens another movie (`drops queued checks once the user opens another movie`) passes against the current implementation, confirming the protection already exists; the captured-token variant broke the existing concurrency test precisely via that spurious session bump. The rationale is now a comment in `check()` so it is not re-introduced.

4gray:
@codex review

Two commits landed since your pass on `5b9339242`: `63ce63845` (e2e selectors) and `ea505a6a9`, which restores provider playback for a downloaded movie — the fix for your first finding. Please re-check the action row on the final head.

4gray:
@greptileai

Re-review requested: your 5/5 was on `5b9339242`, and two commits have landed since — `63ce63845` (e2e selector fix) and `ea505a6a9`, which restores provider playback for a downloaded movie in a single-playlist library. The action row and the check queue's session handling are the areas worth a fresh look.

PR #1361, opened 2026-08-02
refactor(styles): move nav-list partial into ui-styles

4gray:
Stacked on #1360 — please merge that one first. Base is `claude/infallible-leakey-b1707c`, so this PR shows only its own commit.

## Why

`apps/web/src/nav-list.scss` was imported by relative path from five library stylesheets, inverting the intended lib → app direction. #1360 documents this as the one case its two rules **cannot** repair: the file is already inside `web:build`'s own inputs, so `styles:inputs:validate` stays silent, and a lib → app edge cannot be declared because it would make the graph cyclic (`web` already depends on all three libraries). The only fix is to move the partial.

This is not a caching bug today — it is the structural exception the boundaries doc explicitly says to refactor away.

## What

- `git mv apps/web/src/nav-list.scss` → `libs/ui/styles/_nav-list.scss`, with the same `@use`-convention header the sibling partials carry, and `@forward 'nav-list'` added to the `_index.scss` inventory.
- Repointed all five importers. Four now `@use` the canonical partial directly; `workspace-context-category-view.component.scss` goes through the library's own `src/nav-list.scss` entry point (repointed too) instead of an 8-deep path, matching its two siblings in that project.
- **No new declarations needed.** All three consuming projects already declare `"implicitDependencies": ["ui-styles"]`, and `apps/web` imports nothing from `ui/styles` — its only cross-project stylesheet import is `styles.scss` → `libs/ui/components/.../resizable.scss`, which runs app → lib and is already inside the closure.
- No library stylesheet imports from `apps/` any more.

## Docs

- `docs/architecture/nx-workspace-boundaries.md` — the "Shared Stylesheets and Cache Inputs" exception paragraph named the file this PR moves; rewritten to state the rule generically and record that nothing under `libs/` imports from `apps/`.
- `docs/architecture/iptvnator-ui-guidelines.md` and `.codex/skills/iptvnator-ui-design/SKILL.md` — path references.

## Validation

| Check | Result |
| --- | --- |
| `pnpm run styles:inputs:validate` | green — 42 imports across 133 files |
| `pnpm nx build web` | succeeds (only pre-existing budget/CommonJS warnings) |
| `pnpm run skills:validate` | 8 skills OK |
| `pnpm nx affected -t lint` | 14 projects pass |

**Cache wiring proven at the new location:** appending a comment to `libs/ui/styles/_nav-list.scss` made `web:build` re-run (3 of 4 tasks cached, not 4).

**Styling verified by byte-diff rather than by eye.** I built the pre-move state with `--skip-nx-cache` and diffed the whole `dist/apps/web` output against the post-move build: all 232 files identical except `ngsw.json`'s timestamp. Angular inlines component styles into the JS chunks, so identical chunks prove the nav-list CSS is emitted identically for the M3U sidebar, portal rail links, both category views, and the workspace context panel — no rendered difference is possible.

## Release note

Skipped — pure refactor, no user-visible change; hence the `no-release-note` label. (`.scss` under `libs/**` is not auto-exempt from the gate.)

Prettier flags two of the touched files, but those are pre-existing `color-mix` line wraps unrelated to these edits, so they are left alone.

### DIFF
--- a/.codex/skills/iptvnator-ui-design/SKILL.md
+++ b/.codex/skills/iptvnator-ui-design/SKILL.md
@@ -9,7 +9,7 @@ description: Use when changing user-visible Angular UI in IPTVnator, especially 
 
 - Policy: `docs/architecture/iptvnator-ui-guidelines.md`
 - Theme and navigation: `apps/web/src/m3-theme.scss`,
-  `apps/web/src/nav-list.scss`
+  `libs/ui/styles/_nav-list.scss`
 - Channel row: `libs/ui/components/src/lib/channel-list-container/channel-list-item/`
 - Shared EPG timeline/list: `libs/ui/epg/src/lib/epg-timeline/`,
   `libs/ui/epg/src/lib/epg-list-view/`
--- a/docs/architecture/iptvnator-ui-guidelines.md
+++ b/docs/architecture/iptvnator-ui-guidelines.md
@@ -37,7 +37,7 @@ Use it when changing existing views or introducing new list-based UI in the work
 - Shared EPG list styles:
   `libs/ui/epg/src/lib/epg-list-view/epg-list-view.component.scss`
 - Shared list selection style:
-  `apps/web/src/nav-list.scss`
+  `libs/ui/styles/_nav-list.scss`
 - Theme tokens:
   `apps/web/src/m3-theme.scss`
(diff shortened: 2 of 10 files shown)


4gray:
@codex review

PR #1375, opened 2026-08-04
docs(stalker): reconcile the Stalker docs after the API-compatibility series

4gray:
Consistency proofread of the Stalker documentation after the nine-PR API-compatibility series (#1322, #1324, #1334, #1335, #1344, #1348, #1354/#1358, #1364, #1370). Those PRs were written in parallel worktrees and each edited its own section of `docs/architecture/stalker-portal.md` and `CLAUDE.md`, so sections that were correct when written now disagree with each other or with master.

Every claim was verified against the code on master, not against a neighbouring doc.

## Corrected

**`docs/architecture/stalker-portal.md`**

- **Routes** were listed as `/stalker/:id/...`. `createStalkerRoutes()` is mounted under the workspace shell — the real paths are `/workspace/stalker/:id/...`, as `CLAUDE.md` already said.
- **"Simple portal: requests carry only the `mac=` cookie"** contradicted the Request Transport section further down. `stalker.events.ts` and the web-backend proxy both run `buildStalkerIdentityRequestContext()` on *every* request, so a simple portal receives the full STB identity (mac/stb_lang/timezone cookie, MAG UA pair, `SN`); only `Authorization: Bearer` is absent. Also added the explicit "mode is not tied to a URL shape" note the Account Info section already depended on ("a token-enforcing `portal.php` panel is a full portal too").
- The **session facade** is introduced as "a facade over three focused modules" above a list of five.
- The **identity policy** intro still opened with the pre-#1370 "blank fields are not generated", which reads as contradicting the opt-in device-ID derivation documented two subsections below.
- `deriveStalkerDeviceIdsFromMac` was cited by bare filename; **two different modules** are called `stalker-identity.utils.ts` (the function is in `libs/shared/interfaces`).
- The **playback header contract** named two surfaces that apply `ElectronStreamHeadersService`; there are three — `UnifiedLiveTabComponent` (`libs/portal/shared/ui`) applies it for the collection routes' radio player.
- **`get_profile` status decoding** did not record that a bare `{status: 1}` (or refusal text without the status) is a refusal — the behaviour #1354 changed.
- **Backup** listed only `stalkerToken`/`stalkerAccountInfo` as excluded; #1354 added `stalkerSessionIdentity`, `stalkerWatchdogTimeout` and `stalkerTimeslot`, none of which are in `StalkerPlaylistBackupEntry`. Mirrored in `playlist-backup-restore.md`, which carries the same list.

**`CLAUDE.md`**

- There was **no entry at all** for portal mode / endpoint discovery / lazy repair — the largest architectural change of the series (#1344). Added a compact block covering the single predicate, the probe order, `executeStalkerRequest()` as the choke point, lazy-repair-not-migration, and the shared wire-format builders.
- The session facade list was missing `stalker-token-cache.ts` and `stalker-session-store.ts`; `status: 1` still read as plain "blocked" after #1370 split `device-conflict` out of it.

**`apps/stalker-mock-server/README.md` + `docs/architecture/stalker-mock-server.md`**

- Documented the routes `src/main.ts` actually mounts that no doc mentioned: `GET /stalker` (PWA proxy mirror), `GET /stream/gated/:file`, `GET /assets/marketing/poster/*`; plus the `HOST` env var.
- **`/reset` is scoped per MAC now.** Both docs described — and the sample spec demonstrated — the global `POST /reset` that would wipe a sibling spec file's state mid-run. Replaced with the real isolation contract: `OWNED_MACS`, the disjoint `00:1A:79:5F:*` range owned by `self-hosted.e2e.ts` / the `sources-pwa` helpers, `mode: 'serial'`, and the MACs deliberately kept out of `OWNED_MACS`.
- API coverage was missing `get_main_info` (which `CLAUDE.md` already claimed the mock implements) and the `get_genres_itv`/`get_genres_vod` aliases; `create_link` is no longer always a public HLS URL.
- Project-structure tree was missing six files (`auth-store.ts`, `request-mac.ts`, `marketing-poster-url.ts`, and the profile/events/main-info/all-channels handlers).
- Broken in-doc anchor `#two-endpoints-with-different-strictness` (the heading became "Endpoints With Different Strictness" when the third and fourth endpoints landed).
- Stale Playwright `webServer` snippet (2 entries; there are 4).

**`.codex/skills/stalker-portal/SKILL.md`**

- "Keep Stalker request and shape rules in Stalker data access" is no longer accurate: since #1348/#1358/#1370 the wire-format, identity, portal-mode and auth-failure contracts live in `libs/shared/interfaces` precisely because the Electron main process cannot import renderer libs. Added that ownership row and a short portal-mode/session contract paragraph. Still passes `pnpm run skills:validate` (490/500 words).

## Verified as still correct (no change)

Scenario table lists all 11 MACs in `scenarios.ts` including the late `:09` gated-stream and `:0A` static-channel-cmd; watchdog default is 120 s clamped 30–3600 (no `25 s` survives anywhere); "generated MAG-like identity remains a future setting" was fully removed by #1370; the single portal-mode predicate, discovery candidate order, lazy-repair latch key, `enforceMacFormat`-on-strict-endpoint-only, the three plain-text auth bodies and their shared classifier, `STALKER_STB_PROFILE_PARAMS`, the `download-request-headers.ts` allowlist (no Cookie/Authorization), the 15-min playback-context TTL, and every full file path referenced by these docs all resolve. The playback-header (#1335) and temporary-link (#1364) sections do not contradict each other.

## Flagged, deliberately not changed (keeps this PR docs-only)

Two **code comments** carry the same stale claim this PR fixes in prose. Happy to fold them in (they'd need the `no-release-note` label for the `libs/**` one):

- `apps/stalker-mock-server/src/main.ts:107-113` — `isFullPortalUrlShape`'s docblock describes itself as mirroring "the app's full-portal predicates (`isFullStalkerPortal` checks `/stalker_portal/`; import-time normalization checks `/stalker_portal`)". That divergence was unified by #1344 into one predicate; the mock's own URL-shape rule is still correct, only the rationale is stale.
- `libs/portal/stalker/data-access/src/lib/stalker-session.service.ts:55` — class docblock says "Handles handshake authentication for full stalker portals (/stalker_portal/c URLs)", i.e. mode by URL substring.
- `libs/ui/playback/src/lib/web-player-view/electron-stream-headers.service.ts:18-32` — the service docblock lists the same two consumers the doc did; `UnifiedLiveTabComponent` is a third.

Nothing was rewritten from memory: no statement that could not be verified from the code or the roadmap's reference block was touched.

## Test impact

**Docs-only** — all six changed files are `.md`. No runtime code changed, so unit/E2E validation was not required and no release note is needed (the gate's `/\.md$/` exemption covers every file in the diff; no `no-release-note` label required). Validation performed instead:

- `pnpm run skills:validate` → `Validated 8 repository skills.`
- Markdown reviewed for rendering: tables well-formed, all in-doc anchors resolve (`#endpoints-with-different-strictness`, `#endpoints-tolerant-strict-and-the-ministra-host`, `#test-isolation`), and every `apps/`/`libs/`/`docs/` path referenced in the touched docs exists on disk.
- Prettier status is unchanged by this diff (these markdown files were already non-conforming on master; prettier is not run on markdown in CI).


4gray:
Thanks — verified and fixed in 608f487ca. Codex's P2 was correct on all three counts, and the same over-broad claim turned out to already be in `stalker-portal.md` from #1344 ("Every portal API call funnels through `executeStalkerRequest()` … the single choke point"), which is exactly the drift class this PR exists to catch. There is also a fourth exception neither the doc nor the review named.

Confirmed direct `STALKER_REQUEST` callers, each running *below* or *before* what the choke point routes on:

| Caller | Why it must be direct |
|---|---|
| `stalker-auth.api.ts:218/316/367` | `handshake` / `get_profile` / `do_auth` are what `executeStalkerRequest`'s full-portal branch is implemented in terms of — routing them back through it recurses. |
| `stalker-portal-discovery.service.ts:328` | Probes run before a mode exists; the mode is what they determine. |
| `stalker-account-info.service.ts` (`fetchViaProfile` → `refreshAccountProfile`) | A profile request, so it takes the auth layer's exemption. The comment at :131 already said this out loud. |
| `stream-resolver.service.ts:449` | A collection item carrying its own portal coordinates with **no playlist row** — no meta to route or repair with. The `executeStalkerRequest` branch at :439 beside it handles the row-backed case and wins when a row exists, so a repaired endpoint still beats a stale favorite's snapshot. |

Rather than just listing exceptions, the docs now state the boundary, since a bare list goes stale the same way the original claim did:

- the exemption is from the **routing**, not from repair — the callers that can observe a wrong-endpoint failure wire `StalkerPortalRepairService` explicitly;
- anything new that is not auth or discovery belongs on `executeStalkerRequest()`.

Applied to `docs/architecture/stalker-portal.md` (Runtime Architecture step 3), `CLAUDE.md`, and the skill. Skill still passes `pnpm run skills:validate` at 497/500 words.

One related item left deliberately untouched, since this PR is docs-only: `stores/utils/stalker-request.utils.ts:125` opens its docblock with "Single choke point for Stalker API calls" — same overreach, in code. Happy to fold that in (it would need the `no-release-note` label) or leave it for a follow-up.

4gray:
Both reviews above landed on `5d3abc718`, before the fix for Codex's P2 (`608f487ca`) — re-running them over the current head.

@codex review
@greptileai

4gray:
Folded the flagged code comments into this PR (`c242ea633`) — comment-only, no behavior change. The PR is therefore no longer strictly docs-only and now carries the `no-release-note` label for the `libs/**` paths.

Four stale docblocks, all the same class of drift as the prose fixes:

- **`stores/utils/stalker-request.utils.ts`** — "Single choke point for Stalker API calls", the same overreach Codex flagged in the skill. Scoped to catalog/content/playback, with the four direct callers and the rule that the exemption is from the *routing*, not from repair.
- **`stalker-session.service.ts`** — "Handles handshake authentication for full stalker portals (`/stalker_portal/c` URLs)" decided mode by URL substring, which #1344 replaced with the observed, persisted flag. Also corrected "persists tokens during session": since #1354 they outlive the run on the playlist row.
- **`electron-stream-headers.service.ts`** — named two consumers of the single override slot; `UnifiedLiveTabComponent` is a third, for exactly the reason the Stalker live layout is one (renders `AudioPlayerComponent`, never mounts `WebPlayerViewComponent`).
- **`stalker-mock-server/src/main.ts`** — `isFullPortalUrlShape`'s docblock presented the app's three diverging predicates as current. Worth being precise here: the mock's own URL-shape rule is still correct **and deliberate** — a fixture has to decide strictness from the path, because that IS the behavior the app observes. Only the rationale was rewritten, not the rule.

**Validation** (local, since this is no longer docs-only):

- `pnpm run typecheck:ci` — pass (the documented trap: `nx affected -t lint,test` does not typecheck)
- `nx run-many -t lint --projects=portal-stalker-data-access,ui-playback,stalker-mock-server` — pass
- `nx run-many -t test --projects=portal-stalker-data-access,ui-playback,stalker-mock-server` — **91 suites / 915 tests pass**

No new tests: nothing executable changed, so there is no behavior to pin. E2E not run for the same reason.

4gray:
Head is now `c242ea633` (code comments folded in) — please re-review.

@codex review
@greptileai

4gray:
Second P2 is valid too — fixed in `79dd91f0a`. Worth noting it lands on my own compression from `608f487ca`, i.e. an error introduced while fixing the first P2. Two mistakes in one sentence, both confirmed against the code:

**1. Dropped exception.** The skill said "playback calls" route through `executeStalkerRequest()` with auth/discovery/account-profile as the exceptions — omitting the fourth one the prose already listed, `StreamResolverService`'s row-less branch, which issues a `create_link` **playback** request directly. Codex named exactly that.

**2. Repair attribution was wrong.** I wrote that the direct layers "wire repair themselves". Counted `portalRepair` references across the four:

| Caller | refs |
|---|---|
| `stalker-account-info.service.ts` | **6** |
| `stalker-auth.api.ts` | 0 |
| `stalker-session.service.ts` | 0 |
| `stalker-portal-discovery.service.ts` | 0 |

So only one of the four wires it. The real shape, now written down in all four places:

- `fetchViaProfile()` wires `StalkerPortalRepairService` explicitly — its own comment says why (opening the account dialog on a stale endpoint must be able to fix it rather than wait for an unrelated catalog request).
- `StalkerPortalDiscoveryService` is what repair *drives*, so it cannot repair itself.
- `StreamResolverService`'s row-less branch has no playlist to repair.
- **The auth layer needs nothing**, which is the part worth recording: a terminal handshake failure propagates out of the full-portal branch and is caught by whichever `executeStalkerRequest()` call triggered the authentication — which is precisely why "terminal handshake failures" is already one of the documented repair triggers (`stalker-portal-repair.service.ts` matches it via `/handshake failed/i`). It is covered *through* the choke point, not beside it.

Applied to `stalker-portal.md`, `CLAUDE.md`, the `stalker-request.utils.ts` docblock and the skill.

**Validation:** `pnpm run skills:validate` (497/500 words) · `nx run-many -t lint,test --projects=portal-stalker-data-access` → 27 suites / 459 tests pass. Comment/docs only, no executable change.

@codex review
@greptileai

4gray:
Third P2 is valid too — fixed in `653f74dc4`. I wrote "`MOCK_PORT` overrides the mock's port for both the config and the specs"; the second half is right, the first is not.

Verified — **nothing maps `MOCK_PORT` to `PORT` anywhere**:

- `main.ts` reads `process.env.PORT` (default 3210).
- `apps/stalker-mock-server/project.json` pins `PORT: "3210"` in the `serve` *and* `serve-with-watch` target env.
- `MOCK_PORT` is read only by the two Playwright configs and the spec constants (`stalker.e2e.ts`, `self-hosted.e2e.ts`, `sources-pwa.helpers.ts`, `electron-test-fixtures.ts`) — all client side.

So setting `MOCK_PORT` alone points the health check at a port nothing is listening on and the run times out, exactly as described. The doc now says it moves the client side only, and that relocating the Nx-managed server would need it passed through as `PORT`.

I also fixed the **pre-existing** comment in `apps/web-e2e/playwright.config.ts:54` — "Set MOCK_PORT to override the default mock server port (3210)" — since that is where I picked the claim up in the first place. Comment only. Actually wiring `MOCK_PORT` → `PORT` is a tooling change, so I left it as a follow-up rather than smuggling it into a documentation PR; say the word if you'd rather have it here.

**Validation:** `nx run web-e2e:lint` → 0 errors (the 6 warnings are pre-existing in spec files this PR does not touch) · `tsc -p apps/web-e2e/tsconfig.json --noEmit` → clean.

For the record across the three rounds, every Codex finding has been real, and two of the three were on text I wrote while fixing the previous one.

@codex review
@greptileai

4gray:
Fourth P2, valid again — fixed in `db6e479ae`. And again it is text I added in this PR, not pre-existing.

I claimed a simple portal receives the full STB identity "including `SN` when a serial exists", with only `Authorization` absent. The two branches of `dispatchStalkerRequest()` differ by more than the token:

```ts
// full portal
deps.stalkerSession.makeAuthenticatedRequest(...)   // → { url, macAddress, params, token, serialNumber? }

// simple portal
deps.dataService.sendIpcEvent(STALKER_REQUEST, { url, macAddress, params })
```

So on the simple path the shared builder never receives a serial, and therefore emits **neither the `SN` header nor the serial-derived `__cfduid` cookie** — regardless of what the playlist has stored. Corrected the bullet to describe what the builder can derive from a MAC alone, and scoped the identity-policy bullet ("reused for … normal API requests") to the full-portal path, since that is the branch that actually forwards the serial.

**Deliberately not "fixed" in code.** Whether a token-free panel should receive the serial is unproven: no reference portal is known to require it, and the `sn` *parameter* only ever travels on `get_profile`, which a simple portal never calls. Forwarding it would be a transport change on the strength of a guess, which is the opposite of what this PR is for — so it is written down as an open question instead.

Tally so far: four Codex findings, four real, and three of them on prose I wrote while fixing the previous one. That is a useful signal about this particular diff, so I'd rather run one more pass than call it done.

@codex review
@greptileai

4gray:
Fifth P2, valid — fixed in `ca77c859d`, and it caught an **over-correction of my own previous fix**.

In `db6e479ae` I scoped the whole identity bullet to the full-portal path. But that bullet lists five uses, and only the API half is mode-gated:

| Route | Mode-gated? | Why |
|---|---|---|
| Portal API requests | **yes** | serial arrives via `makeAuthenticatedRequest()`, which only the full-portal branch of `dispatchStalkerRequest()` calls |
| Same-origin playback headers | **no** | `buildStalkerExternalPlaybackHeaders()` reads `playlist.stalkerSerialNumber` straight off the row — no mode check anywhere in it |

So a simple-mode playlist holding a real serial **does** send `SN` + `__cfduid` on a portal-owned stream, while its API calls send neither. Both sections now state that asymmetry and point at each other.

---

Five valid findings in a row, four of them on prose I wrote while fixing the previous one. That is a pattern, not bad luck — every time I compress an exact description into one sentence, the qualifier is what falls out. So rather than wait for round six I audited my own diff for absolute claims (`only` / `never` / `every` / `all`) and checked each against the code. Two more of the same kind, now fixed in the same commit:

- **`CLAUDE.md` still said simple portals "send the full STB identity headers the builder produces"** — the exact claim I had just corrected in `stalker-portal.md`. That is the cross-document inconsistency this PR exists to prevent, about to ship inside the PR that prevents it.
- **`CLAUDE.md` credited the mock's `/stalker` mirror with all three shared builders.** It uses the identity builder only: it dispatches in-process, so there is no portal URL to build, and it mirrors the `JsHttpRequest` default by hand. Notably the pre-existing wording in `stalker-portal.md` ("the same shared identity builder", singular) was already precise — my compression was the loose one.

Verified and deliberately left as-is: "every portal path uses the same dispatcher" (both routers and the `/stalker` mirror call `dispatchPortalAction`) and "both transports share the builders" (electron + web-backend, 4 references each).

Docs only; `pnpm run skills:validate` passes.

@codex review
@greptileai

PR #1341, opened 2026-08-02
feat(xtream): catch-up badge for live channels with archive

4gray:
Closes #1128

Live channels with provider catch-up (`tv_archive === 1` with a positive `tv_archive_duration`) now show a small `history` badge:

- **Sidebar channel list**: badge next to the channel name, tooltip shows the archive window ("Catchup available — N days").
- **All-channels grid** (live grids only): dark pill badge on the card, same tooltip. VOD/series grids never render it.
- **Favorites and recently-viewed lists** (portal tabs and global favorites): same badge on Xtream rows.

Follow-ups from @WolfganP's testing feedback (thanks!):

- **Programme-info (i) button is now in the Xtream and Stalker live sidebars too** — previously it existed only in favorites lists.
- **Row actions stack vertically**: favorite on top, (i) below. The stacked column is narrower than the previous single-button row (28px vs 32px), so the extra button costs no horizontal space — channel names actually gain a few pixels.
- **No layout jumps**: the (i) slot is reserved (inert and invisible) while a row's EPG programme hasn't loaded, so the star never shifts.
- Accessibility: the badge status is exposed as visually-hidden text (the icon itself is aria-hidden), and the empty info slot is disabled + aria-hidden.

## Implementation

- Pure helpers `isXtreamCatchupAvailable` / `getXtreamCatchupDays` in `@iptvnator/portal/shared/util` mirror the existing playback-side gates: flag **and** positive window required. String field values from lenient panels are handled.
- `ChannelListItemComponent` gains optional `catchupAvailable`/`catchupDays` inputs (default off) and the reworked action column; `UnifiedFavoriteChannel` carries `tvArchive`/`tvArchiveDuration` through to the shared favorites list.
- Data flows unchanged: both the Electron SQLite path and the PWA API path already carry the archive fields.
- i18n: `CHANNELS.CATCHUP_AVAILABLE` / `CHANNELS.CATCHUP_AVAILABLE_DAYS` added to all 19 locales; the (i) button reuses the existing tooltip key.

## Tests

- Helper spec (flag/window/string/negative cases); badge render on/off, action order, and reserved-slot inertness in `channel-list-item.component.spec.ts`; three grid cases in `grid-list.component.spec.ts`; favorites badge + archive passthrough in `global-favorites-list` / `unified-live-tab` specs.
- E2E: badge visibility asserted on the sidebar row in `xtream-epg.e2e.ts`, green locally in both timezone variants.
[...]

PR #1387, opened 2026-08-08
feat(ui): programme details in the live channel context menu

4gray:
Follow-up to the discussion in #1341: the Xtream and Stalker live sidebars intentionally have no per-row (i) button (the current programme is already shown inline), but that left no way to read a programme description without switching channels. Instead of adding a third button to every row, the programme details are now reachable through the row context menu, consistently across live lists.

## What changed

- **Xtream live sidebar** (`portal-channels-list`) and **Stalker ITV sidebar** (`stalker-live-stream-layout`): right-click now offers **"Show program details"** (opens the existing `EpgItemDescriptionComponent` dialog) whenever the row has a current preview programme.
- **Favorites / recently-viewed lists** (`global-favorites-list`, used by the portal favorites/recent tabs and global favorites): the same item appears when the row has a current programme.
- The context-menu availability gate now keys on what the menu can actually offer (mapping support **or** a current programme) instead of EPG-mapping support alone. Note this is future-proofing, not PWA support: EPG previews are currently populated only where `supportsEpg` (Electron), so in the PWA the maps stay empty and the item does not appear — behavior there is unchanged. If portal EPG previews ever land in the PWA, the menu lights up automatically. "Map EPG" itself stays Electron-gated.
- Reuses the existing `EPG.PROGRAM_DIALOG.SHOW_PROGRAM_DETAILS` key — no new translations needed. Icon is `event_note` to avoid clashing with the M3U "Show details" (channel metadata) item that already uses `info` in the favorites menu.

## Rationale

The sidebar row already shows the programme title, time, and progress inline, and selecting a channel opens the full EPG timeline — a third per-row button would add action density to a high-frequency zapping surface and appear/disappear per row depending on EPG coverage. The context menu gives consistent access everywhere with zero visual noise.

## Tests

- `portal-channels-list` spec: menu offered only when a programme exists (no mapping support in harness), programme lookup, dialog open.
- `stalker-live-stream-layout` spec: programme lookup and dialog open via the preview-programme map.
- `global-favorites-list` spec: enriched-row programme flows into the menu and opens the dialog.
- All three suites green; lint clean; `nx build web` (template typecheck) passes; release note added (`release:notes:validate` passes).

### DIFF
--- a/.changes/ui-program-details-context-menu.md
+++ b/.changes/ui-program-details-context-menu.md
@@ -1,0 +1,9 @@
+---
+type: feature
+area: ui
+---
+
+Right-clicking a live channel now offers "Show program details" in the
+context menu — in the Xtream and Stalker channel sidebars and in the
+favorites and recently-viewed lists — whenever the channel has current
+programme information.
--- a/libs/portal/shared/ui/src/lib/components/global-favorites-list/global-favorites-list.component.html
+++ b/libs/portal/shared/ui/src/lib/components/global-favorites-list/global-favorites-list.component.html
@@ -62,6 +62,18 @@
 ></div>
 
 <mat-menu #channelContextMenu="matMenu">
+    @if (contextMenuChannel()?.currentEpgProgram) {
+        <button
+            mat-menu-item
+            data-test-id="program-details-menu-item"
+            (click)="openProgramDetails()"
+        >
+            <mat-icon>event_note</mat-icon>
+            <span>{{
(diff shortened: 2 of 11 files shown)


4gray:
@codex review

4gray:
@codex review

4gray:
Superseded: instead of hiding programme details behind the context menu, the (i) button now appears in all live lists with the row actions stacked vertically (star on top, info below, reserved slot — no horizontal cost, no layout jumps). Landing as part of #1341 where the row layout was already being reworked.

PR #1418, opened 2026-08-11
refactor(portals): compact credential-free logging for failed Stalker requests

4gray:
## Summary

Failed `STALKER_REQUEST` handlers dumped the **entire axios error object** (config, request internals, keep-alive agent state, full stack) into the main-process console — roughly 100 lines per request against a dead or slow portal. Browsing a dead portal's catalog produced pages of noise.

This PR extracts the existing compact Xtream error formatter into a shared `portal-request-error.util.ts` and uses it from both portal handlers:

- **Stalker** now logs `[STALKER_REQUEST] Failed { action, host, pathname, type, code, status, message }` — one line instead of a full error dump. The redundant pre-throw `[StalkerEvents] HTTP Error:` line is removed; HTTP ≥400 already reaches the catch block and is now logged once, compactly.
- **Xtream** keeps its exact log output, now via the shared util (−49 duplicated lines).

## Safety notes

- Only `host` + `pathname` of the request URL are retained — the query string (Xtream credentials, Stalker session cmd) structurally cannot reach a log line. `redactSensitiveData` stays applied on top at both call sites.
- The throw shapes crossing `ipcRenderer.invoke` are untouched — the renderer's Stalker endpoint discovery keeps its timeout-vs-connection classification.

## Test plan

- New `portal-request-error.util.spec.ts` (6 cases), including the key contract: a URL carrying `username`/`password` query params never leaks into the formatted output.
- `pnpm nx test electron-backend` — green (full suite).
- `pnpm nx lint electron-backend` — clean.
- `tsc --noEmit` over `tsconfig.app.json` — clean.
- `pnpm run release:notes:validate` — 101 notes valid (adds a `type: internal` note).

### DIFF
--- a/.changes/portals-compact-request-error-logs.md
+++ b/.changes/portals-compact-request-error-logs.md
@@ -1,0 +1,9 @@
+---
+type: internal
+area: portals
+---
+
+Failed Stalker portal requests now log a compact, credential-free summary
+(action, host, error code, HTTP status) in the desktop backend instead of
+dumping the full multi-page network error object, matching the existing
+Xtream request logging.
--- a/apps/electron-backend/src/app/events/portal-request-error.util.spec.ts
+++ b/apps/electron-backend/src/app/events/portal-request-error.util.spec.ts
@@ -1,0 +1,98 @@
+import { AxiosError } from 'axios';
+import { formatPortalRequestError } from './portal-request-error.util';
+
+describe('formatPortalRequestError', () => {
+    const xtreamUrl =
+        'http://provider.example:8080/player_api.php?username=user&password=secret&action=get_vod_streams';
+    const stalkerUrl =
+        'http://portal.example/stalker_portal/server/load.php?action=get_ordered_list&type=vod&p=1&JsHttpRequest=1-xml';
+
+    it('keeps only host and pathname of the request URL, never the query', () => {
+        const formatted = formatPortalRequestError(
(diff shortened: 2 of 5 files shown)


4gray:
@codex review

4gray:
Fixed in 4bd11f263 — thanks, this one was real.

Verified the leak empirically before fixing: `formatPortalRequestError(err, 'http://user:sup3rsecret@')` produced `pathname: "http://user:sup3rsecret@"`, and passing that through `redactSensitiveData` returned it **byte-identical**. The reason is exactly as described: `redactUrl` only sanitizes userinfo of URLs `new URL()` can parse, and `http://user:secret@` (no host) throws, after which every remaining redaction pass declines it — `redactUrlStrings` re-parses and gives up, `redactEmbeddedSensitivePairs` doesn't match because its key alternation has `username`, not `user`.

Fix: the unparseable branch now emits a fixed `[unparseable-url]` marker instead of the raw string. Regression test added, and confirmed it fails against the old fallback (2 failed / 5 passed) and passes with the fix (7/7).

Also added a contract note on the util: callers must pass a portal **API endpoint**, since Xtream stream URLs embed credentials in the path itself (`/live/<user>/<pass>/id.ts`) and host+pathname would retain those. Not reachable from either current caller (both hit fixed API paths), but the util is shared and generically named now.

@codex review

4gray:
@greptileai

PR #1420, opened 2026-08-11
feat(playlist): open recognized M3U movies in the VOD detail view

4gray:
## Summary

- M3U playlists routinely mix live channels with movie files ("Movies"/"VOD" groups, Xtream-derived exports); those entries currently play against a useless empty EPG zone. This PR recognizes them and opens the same two-state VOD detail experience the Xtream/Stalker portals use, fed by TMDB metadata. Works in both Electron and the PWA.
- **Watch-first**: clicking a channel keeps M3U zapping semantics — playback starts immediately with VOD seek semantics, the About block (plot, cast chips, director, genres, rating, runtime) renders below the player where the EPG zone would be, and Esc/close reveals the full Browse hero with poster and backdrop.
- Recognition is a **synchronous URL-shape heuristic**, so the layout never jumps when TMDB answers: a movie container extension (`mkv`/`mp4`/`avi`/… — deliberately not `ts`/`m3u8`/`m4s`, which deliver live streams, and not `mpd`, which has its own DASH path) or an Xtream-style `/movie|movies|vod/` path segment. Radio, DASH, `/series/` paths, and names carrying an episode marker (`S01E02`, `1x02`, "2 серия", "Season 3") fail toward today's live layout.

## Changes

- `libs/shared/m3u-utils`: new `isLikelyM3uMovie` / `hasEpisodeMarker` detection util (+49 spec cases incl. Cyrillic markers and the documented "Star Wars: Episode 1" cost).
- `libs/playlist/m3u/feature-player`: new `m3u-vod-detail/` host — `PortalDetailShellComponent` + `PortalInlinePlayerComponent` reused as-is; the parent's `embeddedPlayback()` is already a `ResolvedPortalPlayback`, so the playback glue is a spread with `isLive: false` and the resolved title/poster. `M3uVodMetadataService` wraps `TmdbEnrichmentService.enrichMovie` (raw channel name + `releaseTagYear` — the resolver already normalizes titles and caches verdicts) with a staleness guard keyed on channel id.
- `VideoPlayerComponent`: a computed gate (`showMovieDetail`) and a template branch; all new logic lives outside the baselined file. External MPV/VLC launch path untouched — those users get the Browse layout.
- Settings: new root-level `Settings.m3uVodDetails` (default **on**), checkbox in Settings → Metadata (TMDB), visible only while TMDB is enabled. The whole feature is inert unless TMDB enrichment (already opt-in) is active, so the effective rollout is conservative.
- i18n: 2 new keys translated across all 18 non-English locales via the `i18n-fill` pipeline.
- Docs: new "Movie Recognition (VOD Detail View)" section in `docs/architecture/m3u-playlist-module.md`, consumer notes in `docs/architecture/tmdb-metadata-enrichment.md`, CLAUDE.md updated. Release note added under `.changes/`.

## Out of scope (follow-ups)

Series episodes (detection skips them), playback-position persistence for M3U movies, clickable cast chips (no `actor/:personId` route in the M3U tree), downloads, and an e2e with a TMDB mock.

## Testing

- [x] `m3u-utils` 141/141 (49 new detection cases)
- [x] `playlist-m3u-feature-player` 75/75 (new: metadata service 9, detail host 7, recognition-gate 5 in a dedicated spec — the main video-player spec sits at the max-lines test budget)
- [x] web settings specs 158/158 (3 new TMDB-section cases), `services` settings-store 20/20
- [x] `nx build web`, lint across the 5 touched projects, `styles:inputs:validate`, `release:notes:validate`
- [ ] Manual TMDB run needs a real API key (dev builds carry the empty placeholder): enable TMDB in settings and click any `.mkv`/`.mp4` playlist entry


4gray:
Both P1 findings were real and are fixed in e8f9592ec:

**Saved volume dropped** (Greptile + Codex) — `PortalInlinePlayerComponent` had no volume input and `WebPlayerViewComponent.volume` defaults to `1`; the HTML5 player only *writes* to `localStorage`, so the persisted value reaches it exclusively through that input. Added an optional `volume` input to the inline player (default `1`, so the Xtream/Stalker hosts are byte-identical) and threaded `VideoPlayerComponent.volume()` through the detail host.

**Metadata updates restarting playback** (Codex) — confirmed against `createWebPlayerApplicationState`: `sourceRevision` reads `playback()` unconditionally, so any new payload identity mints a new source revision and re-applies the source. `inlinePlayback()` no longer reads any TMDB signal — it is `{...parentPayload, isLive: false}` and changes identity only when the parent payload does. The resolved title/poster stay in the hero/About presentation.

Regression coverage added: payload identity is asserted stable across enrichment arrival (and still changing on channel switch), plus a volume-forwarding test. Suites: playlist-m3u-feature-player 78/78, ui-playback 975/975 (three `app-web-player-view` stubs needed the new input), xtream+stalker feature 36 suites green.

Also note the earlier `Build on macos arm64` failure was infra, not code: `RequestError: self-signed certificate` from `got` while electron-builder downloaded 7zip, after the DMG had already been built.

@greptileai

4gray:
Rechecked the one red E2E job before touching anything: `downloads.e2e.ts:411` ("keeps global and scoped libraries truthful while filters, search, grouping, and scrolling stay functional") timed out on the `?q=` URL-sync poll. Same test, same 115-passed/1-failed shape, failed on **master** in run [31568352129](https://github.com/4gray/iptvnator/actions/runs/31568352129) at 36c2867d — a pre-existing flake, not this PR: the downloads route owns `q` through its own `queryParamSignal` and never touches the inline player or the M3U branch.

Re-triggering a scored review on the new head e8f9592ec (the summary above still points at 6ec681618).

@greptileai

4gray:
Fixed in 2930729e8 — and the finding is worth stating precisely, because it is **not** a regression from this PR: `VideoPlayerComponent.volume` was a one-shot constructor read of the shared `volume` localStorage key, and the engines persist their own changes straight to that key without calling back. Since `web-player-view` renders `@for (application of renderedApplications(); track application.token)`, every source revision (i.e. every channel switch) creates a fresh engine instance that re-reads the input — so the live M3U layout already snapped playback back to the page-open volume after the user adjusted it. The movie detail path inherited that faithfully.

Fix is one signal in the parent: `volume` is now a `linkedSignal` sourced on the active channel id, so each new player instance reads the current bus value. Remote-control writes still `set()` it and store the same value, so both agree.

While covering it I hit a defect in my own first draft that the new tests caught: `Number(localStorage.getItem('volume'))` maps **both** `null` and `''` to `0`, so a first run with nothing stored would have started muted. The reader now rejects the empty cases before `Number()` and clamps out-of-range values to 1, with `'0'` still honored as real silence.

Coverage: bus re-read across a channel switch, the empty/invalid/out-of-range table, first-run default, and stored silence. playlist-m3u-feature-player 86/86, lint clean.

@greptileai

4gray:
The whitespace case is already in the table — `it.each(['not-a-number', '', '  ', '5', '-1'])` at [video-player-movie-gate.spec.ts:333](https://github.com/4gray/iptvnator/blob/2930729e8/libs/playlist/m3u/feature-player/src/lib/video-player/video-player-movie-gate.spec.ts#L333), so `'   '` falling through to `1` is pinned rather than incidental. And yes, the `?? null` source is deliberate: before a channel is selected the source is a stable `null`, so the computation runs once instead of re-firing on playlist load.

Note the scored summary above (4/5) is stamped 16:36:29Z, which predates the volume fix commit 2930729e8 (16:41:35Z) — it graded e8f9592ec. Requesting a scored pass on the actual head.

CI is now fully green (19/19). For the record, all four failures seen on this branch were infrastructure or pre-existing flakes, each confirmed on master or by rerun:
- `Build on macos arm64` — `RequestError: self-signed certificate` fetching 7zip; green on rerun
- `Build on linux x64 (system)` — `RequestError: socket hang up` during `make`; green on rerun
- `Build on linux x64 (portable)` — packaged frame-copy smoke never got a frame under llvmpipe; identical failure on master in run 31567281884
- `Electron E2E on ubuntu-latest` — `downloads.e2e.ts:411` `?q=` poll timeout; identical failure on master in run 31568352129

@greptileai

4gray:
Checked this one against Angular's implementation rather than taking it at face value, and **the premise does not hold** — though I've taken the suggested source anyway, for a different reason.

`linkedSignal` does not gate its computation on the derived source *value*. In `producerRecomputeValue` ([`_untracked-chunk.mjs`](https://github.com/angular/angular/blob/main/packages/core/primitives/signals/src/linked_signal.ts), Angular 21.2.17) the node re-evaluates `node.source()` and then calls `node.computation(newSourceValue, prev)` **unconditionally** — the previous source value is only handed to the computation as `prev`, never compared to skip work. So the node re-runs whenever the source expression's dependencies invalidate, which selecting a different `Channel` object always does, equal ids or not.

Verified empirically too: with the original `source: () => this.activeChannel()?.id ?? null` restored and the new same-id test in place, the switch still picks up the newly stored volume (87/87 green either way). So the described stale-volume path was not reachable.

That said, I kept `source: () => this.activeChannel()` in 3170b21e7: it states the intent ("re-read per selection") without requiring the reader to know that subtlety, and it costs one idempotent localStorage read. The code comment now explains the real mechanism instead of the id-collision reasoning, and the same-id test stays as a behavioural guard — it pins what users actually experience regardless of what the source derives.

Worth flagging the underlying fact separately: `createChannel` really does fall back to `params.url` for the id, so duplicate entries sharing an id are a real shape in this codebase. It just isn't what this signal depends on.

@greptileai

4gray:
@codex review

4gray:
Both Codex findings addressed in d75e656de.

**P2 — volume on remount.** Real, and specific to this feature: Browse → Play remounts the engine without a channel change, and the volume signal only recomputed per channel, so the remounted player got the pre-adjustment value. The detail host now emits `playbackStarted` and the parent re-reads the shared bus — keeping the parent the single owner rather than letting the host read localStorage behind its back.

**P1 — E2E coverage.** Added `apps/web-e2e/src/m3u-movie-details.e2e.ts`, driving the real composition with a routed TMDB mock:
- recognized entry opens the detail view (player mounted, no `app-epg-timeline`), TMDB overview/genre/director patch in asynchronously, and a live channel in the same playlist keeps the classic layout;
- a volume persisted before the movie is opened reaches the player, and survives Escape → Play.

The volume test is a real guard, not decoration: with the `(playbackStarted)` wiring removed it fails at exactly the post-remount assertion, and passes with it.

Two deliberate choices worth naming. The test never decodes media — the claims are layout, metadata and volume, so a decodable fixture would only add flakiness. And it drives volume through the shared localStorage bus rather than poking `video.volume` and waiting for the engine to persist: that engine-internal wiring already has unit coverage, and depending on it made the test flaky across runs.

Local: 89/89 unit in playlist-m3u-feature-player, 2/2 new E2E on chromium, lint clean across the touched projects.

4gray:
@codex review

4gray:
Fixed in bb8251186, with one correction to the premise.

The guard now keys on the full lookup identity (`id` + title) instead of the id alone. That is the right shape regardless: the request is built from `channel.name`, so deduping on the id assumed an id determines the name, and nothing guarantees that.

But the specific path described is **not reachable today**: M3U import assigns `id: createRandomId()` per entry ([playlist.utils.ts:400](https://github.com/4gray/iptvnator/blob/bb8251186/libs/shared/m3u-utils/src/lib/playlist.utils.ts#L400)), and the `createChannel` helper carrying the URL fallback has **no production callers at all** — the only `createChannel` functions in the workspace are same-named test helpers. Duplicate `tvg-id` values also cannot collide here, since the id is not derived from `tvg-id`. So this was a latent coupling rather than a live defect, and I fixed it as such.

Coverage: two entries sharing an id with different titles now each get their own lookup (and the in-flight response for the first cannot land on the second). 90/90 unit, lint clean.

@codex review

4gray:
Confirmed and fixed at the root in d0d29567b.

Verified the mechanism rather than just the symptom: `selectedPlayer` falls back to `this.settings()?.player ?? VideoPlayer.VideoJs`, and `settings` is `toSignal(storage.get(...))` — an async IndexedDB read **with no initial value**. Since `selectedPlayer` feeds the application token, anyone whose saved player is not Video.js gets a first application on the wrong engine and a swap when the read lands. On the M3U live layout that never showed because the branch passes an explicit `playerOverride`; the detail path goes through `PortalInlinePlayerComponent`, which passes none — so this is a real regression of this feature.

I fixed the cause instead of threading another override input: `SettingsStore` already backs `showCaptions` in this very component and holds the player synchronously (the `/workspace` route awaits its initial load), so it is now the fallback ahead of the hard-coded default. Host overrides still win, and an unhydrated store yields exactly the Video.js default that was hard-coded before — never worse than today. This also removes the same transient wrong-engine mount from the Xtream/Stalker inline players, where it was pre-existing.

Coverage: new `web-player-view.component.player-fallback.spec.ts` pins all three branches with the storage read left pending. It fails without the fix (`Expected "html5", Received "videojs"`). Suites: ui-playback 978/978, playlist-m3u-feature-player 90/90, the M3U movie E2E 2/2, lint clean.

@codex review

4gray:
Landed differently than my previous comment said — with two corrections I owe you.

**The fix.** I first patched the cause inside `WebPlayerViewComponent` (synchronous `SettingsStore` fallback ahead of the pending IndexedDB read). CI rejected it: that file sits at exactly the 400-line hard cap, so my one line made it 401, and extracting a helper only broke even once the import was counted. A real split of that component is its own piece of work and does not belong in a feature PR. So I took the first option from your comment instead — thread the already-resolved selection through the chain: `PortalInlinePlayerComponent` gained an optional `playerOverride` (null keeps the portals exactly as they are), and `VideoPlayerComponent → M3uVodDetailComponent` forward the same value the live layout already passes. 5b4079726.

Trade-off stated plainly: this fixes the regression this PR introduced, and leaves the portals' own transient wrong-engine mount (pre-existing, same root cause) for a change that can afford to split that file.

**The test.** My first attempt asserted `app-html-video-player` present / `app-vjs-player` absent — and I checked it against the unfixed code: **it passed**. Playwright locators poll, and the defect is a mount that is corrected a tick later, so both paths converge on the same final DOM. The guard is now a MutationObserver installed before activation that records every engine ever attached, asserted to be exactly `['app-html-video-player']`. That one does fail without the fix (`Received + 1`) and passes with it.

Suites: playlist-m3u-feature-player 91/91, ui-playback 975/975 (three inline-player stubs needed the new input), M3U E2E 2/2 chromium, lint clean across the three touched projects — verified by exit status this time, after taking a truncated `run-many` tail for success last round and shipping the red lint above.

@codex review

4gray:
Codex is clean on the current head (`5b4079726e`) and CI is 19/19 green.

Greptile's scored summary still points at `2930729e8` — four commits back — and its single blocking finding there was the `linkedSignal` one it retracted in writing after I showed `producerRecomputeValue` calls `computation()` unconditionally. Everything since (volume on remount, the E2E suite, the lookup-identity key, the engine override) is unreviewed by it. Requesting a final pass on the actual head.

@greptileai

PR #1422, opened 2026-08-12
feat(portals): add "View in portal" action to inline collection details

4gray:
## Problem

A movie or series opened from the dashboard (hero, Continue Watching, favorites rails), from **Global favorites** / **Recently viewed**, or from a portal's own favorites/recent tabs renders full-width — there is no category sidebar, and no way back to where the title actually lives in its portal. The Download Manager's focused offline detail already solves this with a `View in portal` action; the collection details had no equivalent.

## Solution

A shared `app-view-in-portal-action` renders as its own row **below** the main hero actions (Play / Favorite / Download), styled like the Download Manager's button.

Visibility is decided by dependency injection rather than URL sniffing: the action renders only when a host provides the new `VIEW_IN_PORTAL_HANDOFF` token, and the only providers are the two collection-detail wrappers (`XtreamCollectionDetailComponent`, `StalkerCollectionDetailComponent`). Those wrappers exist exclusively where a detail is opened outside category context, so a router-mounted portal detail never shows the button and needs no opt-out.

One line added to each of the four `appDetailActions` templates therefore covers every entry point at once — global routes, portal-scoped favorites/recent, and all the dashboard handoffs that funnel into them.

## Navigation contract

Targets come from the new `getUnifiedCollectionDetailNavigation()`. Unlike the existing `getUnifiedCollectionNavigation`, it **never degrades** to a category- or section-only route: an Xtream item without a resolvable category and a positive item id keeps the action hidden instead of promising a jump to the title and landing the user in a list.

Two deliberate differences from the download handoff:

- It does **not** pass `detailPresentation: 'provider-only'`. The item exists in the provider catalog, so the full normal detail (downloads included) is what's wanted.
- Stalker carries `stalkerReturnTo`, so the portal detail's back affordance returns to the originating collection rather than the category grid.

## Verification

Verified end-to-end in a built Electron app over CDP against a real multi-source setup:

| Scenario | Result |
| --- | --- |
| Global favorites → Xtream movie | `/vod/44906/2115519`, sidebar category `\|RU\| FILM 2024/2026` selected |
| Global recent → Xtream movie | `/vod/44906/2115520`, category selected |
| Portal-scoped favorites, same playlist | `/vod/44906/2115519`, category selected |
| Stalker recent → movie | `/stalker/…/vod/2`, sidebar `Наше кино`, detail open via `openStalkerItem` |
| Router-mounted portal detail | no button |

The same-playlist case was the one real risk: both wrappers restore a store snapshot in `ngOnDestroy`, which could have raced the target route's re-sync. The router destroys the deactivated component before `NavigationEnd`, so `syncRouteState` wins — confirmed above rather than assumed.

## Tests

15 new tests: the shared component's spec (token absent / unavailable / click / tooltip), the navigation builder's spec (full routes, the non-degrading `null` cases, uid-tail fallback, Stalker state + `returnTo`, live/M3U rejection), and two tests in each host spec covering token reachability and the emitted navigation.

`pnpm nx run-many --target=test --projects=components,portal-shared-util,portal-xtream-feature,portal-stalker-feature,ui-playback` → **975 passed**. Lint clean, `pnpm nx build web` passes, `pnpm run release:notes:validate` passes.

Translations for both new keys are filled in for all 19 locales via the repo's i18n merger.

## Notes for review

- `libs/portal/stalker/feature/src/lib/stalker-collection-detail.component.ts` is now 529 lines against the repo's 400-line production limit. It was already over and already baselined in `tools/eslint/max-lines-baseline.mjs`; this PR adds ~22 lines. A follow-up split is worth doing so it can drop off that list — happy to do it separately rather than widen this diff.
- `embedded-mpv-session-controller.spec.ts` failed locally on two runs, but with a **different pair of tests each time**, which points at a timing flake rather than a regression: the spec flushes async chains with real `setTimeout(…, 0)` and shims `requestAnimationFrame` onto the same timer, so under parallel load one macrotask tick isn't always enough. It is untouched by this branch (the only `ui-playback` changes here are one template line and one import), and the master commit merged underneath it (#1405) touches no file in `libs/ui/playback`. I did not finish an isolated run on a clean `master` to prove it, so flagging rather than claiming.

### DIFF
--- a/.changes/portals-collection-view-in-portal.md
+++ b/.changes/portals-collection-view-in-portal.md
@@ -1,0 +1,9 @@
+---
+type: feature
+area: portals
+---
+
+Movie and series details opened from the dashboard, global favorites, history,
+or a portal's favorites/recent tabs now offer a "View in portal" button below
+the main actions. It jumps straight to the title inside its own portal, with
+the matching category selected in the sidebar.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1198,7 +1198,9 @@ engine` (restart required) or
   form a metadata-only target; unproven episode and legacy-movie handoffs stay
   unavailable. The normal detail uses one-shot `provider-only` presentation:
   it exposes provider content/playback it can resolve while hiding
-  Offline/local/download actions.
+  Offline/local/download actions. A second, independent `View in portal`
+  bridge exists for inline collection details — see **Collection Detail
+  Portal Handoff** below; it deliberately does NOT use `provider-only`.
 - Download rows and local files survive source deletion. The global offline
   library remains visible with no playlists; only provider handoff is disabled
   until the source exists again.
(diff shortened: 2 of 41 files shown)


4gray:
Both review findings addressed in fa97aa1ad.

**Codex P1 — lazy Stalker VOD series routed to `/series` (valid, fixed).** Traced it end to end rather than taking it on faith: `extractStalkerItemType()` (`stalker-item.normalizer.ts:96`) returns `series` whenever `is_series` is set, regardless of the real category, so a lazy Ministra VOD item reached the builder as `contentType: 'series'`. `StalkerCatalogFacadeService.selectItem()` gates `needsSeriesFetch` on `contentType() === 'vod'`, so the `/series` route did open a detail that could never load its episodes. `getUnifiedCollectionDetailNavigation()` now mirrors `StalkerCollectionDetailComponent.resolveDetailMode()` for the section and `resolveSelectedCategory()` for the virtual-category normalization, exactly as suggested. Embedded `series[]` snapshots take the same path.

Three regression tests were added, and I verified each one **fails on the previous behavior** (reverted the logic, got exactly 3 failures / 9 passes, then restored) so they cannot pass vacuously.

Worth noting for scope: the pre-existing sibling `getUnifiedCollectionNavigation()` passes `item.contentType` the same way, but it is unreachable for this case — `onGridItemSelected()` routes every non-live Xtream/Stalker item through `getGlobalCollectionDetailNavigation()` first. Left untouched rather than silently changing another path's behavior in this PR.

**Greptile P2 — CLAUDE.md (valid, fixed).** Fair: CLAUDE.md described `View in portal` as download-manager-only, so a future reader would not know a second bridge exists. Added a **Collection Detail Portal Handoff** section covering DI gating, the non-degrading target builder, the Stalker section rule above (explicitly warning against simplifying it to `item.contentType`), and the deliberate absence of `provider-only`, plus a cross-reference from the download bullet. `AGENTS.md` mirrors only the process sections, so no sync was needed.

Full suite re-run after the fix: `components` 169, `portal-shared-util` 199, `portal-xtream-feature` 384, `portal-stalker-feature` 271 — all green; lint clean.

One correction to my own PR description: `embedded-mpv-session-controller` has now failed locally on a **third** run with a different test again (this time in `embedded-mpv-session-controller.lifecycle.spec.ts`), while the CI job passed on this branch. That settles it as a pre-existing timing flake in those specs rather than anything from this branch; tracked separately.

4gray:
@codex review

4gray:
@greptileai

4gray:
@codex review

Issue #74, opened 2021-08-24
Referrer/Origin-locked streams won't work


LaneSh4d0w:
Due to some factors, some links might be locked using where they came from (referrer and/or origin). Some links, because of that, can be falsely given a 403 when trying to load them.

4gray:
How are user-agent and referer defined in the link? At the moment iptvnator supports two formats, see https://github.com/4gray/iptvnator/issues/57

LaneSh4d0w:
> How are user-agent and referer defined in the link? At the moment iptvnator supports two formats, see https://github.com/4gray/iptvnator/issues/57 

Defined as external VLC-specific options.

#EXTVLCOPT:http-referrer=https://ssh101.com/ as an example.

Maybe you can translate this to make it work in the Kodi way.

If you need some things :

http-user-agent = User-Agent
http-referrer = Referer
Origin however isn't supported in VLC (or I haven't found yet.

4gray:
Took a random stream from ssh101, in this way it works fine for me: 
`#EXTM3U
#EXTINF:-1 tvg-name="test" tvg-id="test",Test
#EXTVLCOPT:http-referrer=https://ssh101.com/
#EXTVLCOPT:http-user-agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_5)
https://tna5.bozztv.com/crotvhd/index.m3u8`

rossbeck:
Was it ever implemented to let us add custom referers to playlists? I can see where it will let me enter a User agent but can't find the referer section. Is the solution to import a file with it already integrated rather than just adding a playlist via URL and providing the details later? Thanks

Issue #76, opened 2021-09-11
Support mp4 for movies?


izoodeh:
the media could not be loaded either because the server or network failed or because the format is not supported.

izoodeh:
the vod system doesn't work in the app because the html player is in an infinite load and the js player doesn't support vod.

how do i add support for vod, movies and tv show?

https://i.imgur.com/wsSe239.png

analyserdmz:
That would be interesting. Voting for it..

Issue #75, opened 2021-09-04
Suggestion: add an X which clears all text in the search bar.


what-a:
Suggestion: Add an "X" to the right hand side of the search field. When a user clicks on the "X" it removes all text in the search field. Google has a similar feature on their web search.  Below is where the X would appear. 

![image](https://user-images.githubusercontent.com/88639906/132103574-5809034d-205f-41f2-b02f-b5aad1f8e412.png)

PR #1424, opened 2026-08-12
fix(portals): time out and fast-fail PWA proxy requests to dead hosts

4gray:
## The problem

`apps/web-backend` proxied every provider call through a bare `axios.get()` with no `timeout`. A provider that accepts a connection and then goes silent held the request until the OS gave up on the TCP connection — minutes, not the 15/30 s budget the Electron handlers use — and the PWA had no way to cancel it.

The connectivity guard (#1421) was deliberately left out of the PWA for exactly that reason: a breaker keyed on "the endpoint did not answer" is useless until *not answering* is bounded.

## What changed

**Timeouts**, matched to the Electron handlers: Xtream 30 s, Stalker 15 s (30 s for `create_link`), `/parse` 30 s, `/parse-xml` 30 s.

These are safe for large downloads, which was verified rather than assumed. On axios' default transport `timeout` is not a wall-clock deadline: it bounds the time to response headers, then continues as the socket's inactivity timeout for the body. A multi-megabyte XMLTV file that keeps delivering bytes is never cut off — only a stalled one is.

**The breaker**, now shared. `libs/shared/host-health` (`scope:shared` / `domain:shared-runtime` / `type:util`) holds the state machine, failure classification and redirect-attribution helpers — no transport, logger or singleton. Electron keeps its process-wide instance and every existing call site is unchanged; the web backend takes one through `WebBackendAppOptions.hostGuard`, so specs drive it with a clock they own.

It guards `/xtream` and `/stalker` **only**. Playlist and XMLTV downloads keep the timeout but no breaker, matching Electron: a download is one request rather than a catalog fan-out, it is usually the direct result of the user asking for it, and it can outlive the half-open trial window.

**`resetHostConnectivityGuard()` no longer no-ops in the PWA.** The breaker lives in the backend process, so it travels to a new `POST /connectivity-guard/reset`, which reads only the origin and never logs the credential-bearing URL. `skipConnectionGuard` now survives the PWA transport too, so Stalker endpoint discovery keeps the exemption it has on the desktop instead of tripping the breaker with its own probes.

**The Stalker fast-fail needs one extra step.** A refusal keeps each route's HTTP 200 `{message, status}` envelope — but `PwaService.forwardStalkerRequest` turns that envelope into `Error('HTTP Error 502: …')` *and* sets `.status`, and `getStalkerRequestErrorStatus` reads both as "the endpoint answered", which would send discovery walking every candidate and fire lazy repair at a host just declared dead. A prior branch keyed on the shared `isHostConnectivityFastFailMessage` rethrows it bare instead. (#1421 exempted guard messages in the *other* two classifiers; the status path is a separate way to lose the same classification.)

## Review

Nine rounds. Greptile 5/5; Codex found a real defect in most rounds, ending docs-only. The ones worth knowing about:

- **The Xtream breaker never opened at all.** The route passes credentials through axios `params`, so the sent URL always carries a query the baseline does not; comparing whole URLs made every ordinary failure look like a post-redirect failure and credit the endpoint. Now compares origin + path.
- **The failed redirect hop is in a different field per transport** — `config.url` for Electron's per-hop requests, `request._currentUrl` for the web backend's following transport. Verified against axios 1.19.0 with a live 302-to-dead-port server. The shared helper reads both.
- **`StubHttpClient` threw bare `Error`s**, so the guard's redirect logic saw neither field and silently did nothing: 57 tests passed over a breaker that could not open. The stub now shapes rejections like axios, including the appended query — with that alone, four existing tests fail against the old comparison.
- Exempt discovery probes must **report** (`countFailures: false`), not be skipped: this route sets no `validateStatus`, so axios rejects every non-2xx *with* `error.response`, and dropping those is what lets the breaker open mid-discovery.
- A hostname that stops resolving is counted; a genuine SSRF policy refusal only releases the half-open slot.

## Validation

`shared-host-health` 41, `web-backend` 58, `web` 281, `services` 368, `electron-backend` 1680; lint, build, coverage policy and release notes clean. Every guard behaviour added here was mutation-checked — the fix reverted, the specific test confirmed failing.

## Notes

- CodeQL alerts #74/#75/#76 were dismissed as pre-existing: CodeQL is green on master with identical code at all four proxy sites, and they surfaced here only because this PR touched three of the `.get()` call lines. The real gap behind them — the web backend not re-validating redirect hops — is tracked in #1436.
- `web-backend-app.ts` was already baselined for `max-lines`, so the new logic lives in `host-guard.ts` and spec fixtures moved to `web-backend-app.spec-helpers.ts`.
- Docs: `docs/architecture/host-connectivity-guard.md` (ownership, scope, the transport-specific redirect contract) and `CLAUDE.md`.


4gray:
Fixed in `1587c6fec`.

**Redirect failures poison source host** — valid, and the fix went a bit wider than the report. The same defect exists in the Electron half by a different route: `requestWithValidatedRedirects` reissues each hop as its own request, so a dead hop fails against that hop while the guard token still names the original portal. Both halves ship in this PR, so fixing only the one that was flagged would have left the identical bug in the other.

The fix is in the shared classifier rather than at either call site. `classifyHostRequestFailure(error, expectedHost)` now downgrades a host-level failure to `inconclusive` when the transport names a different destination — `request._currentUrl` for follow-redirects (web backend), `config.url` for the per-hop Electron transport. An error that names nothing, such as a plain timeout, still counts: refusing those would disable the breaker for exactly the case it exists to catch. The parameter is optional, so the classification stays backwards compatible.

Regression coverage: four classifier cases in `host-connectivity-guard.spec.ts` (redirect via `_currentUrl`, redirect via `config.url`, same-host failure still counts, unattributed failure still counts) and an end-to-end one on `/xtream` asserting a provider whose redirect destination is dead is never fast-failed. Verified non-vacuous — reverting the fix fails exactly that route test.

Also in this push: `shared-host-health` registered in `tools/coverage/coverage-policy.json` (Tier A), which was the failing **Unit Tests and Typechecks** check — the repo requires every project with a test target to declare a tier.

---

**On the two CodeQL SSRF alerts** (`web-backend-app.ts:378`, `:604`) — not addressed here, deliberately, and I'd like a second opinion before anything changes.

They are pre-existing, not introduced by this PR. All four proxy request sites in that file carry the same `// codeql[js/request-forgery]` comment and the same unvalidated-redirect behavior. CodeQL is green on master with identical code. Only two got reported here, and they are exactly the two whose `.get(` call line this PR modified — CodeQL's PR check reports alerts located on changed lines, so touching the line resurfaced a pre-existing finding. The other two sites (`/xtream`, `/parse-xml`) were filtered as pre-existing because I only changed lines inside their options object.

That said, the underlying concern is real and adjacent to the bug above: **the web backend does not re-validate redirect hops.** `validateProviderUrl` runs only at `/provider-targets` registration, then axios follows redirects freely — so a registered public URL can 302 to a private or link-local address and the proxy will follow it. Electron does not have this gap; `requestWithValidatedRedirects` validates every hop.

I have not fixed that here because it changes request behavior in security-sensitive code, needs the `WebBackendHttpClient` contract to model 3xx, and deserves its own PR and review rather than riding along in a timeout change. Options, for @4gray to pick:

1. Dismiss the two alerts as pre-existing (they are unchanged in substance from master) and open a follow-up issue for redirect validation.
2. I close the gap in this PR — likely via `beforeRedirect` for the synchronous checks (scheme, credentials, literal private/reserved IPs), which covers the metadata-endpoint case but not DNS rebinding.
3. I open a separate PR that ports `requestWithValidatedRedirects`-equivalent hop validation to the web backend properly.

My recommendation is 1 + 3: keep this PR to what it is, and fix redirect validation properly where it can be reviewed on its own merits.

@greptileai

4gray:
Rebased onto current master (`e78fed46a`).

Worth recording why, because the symptom was misleading: after the previous push **no CI ran at all** — zero workflow runs and zero check-runs for that SHA, for ~55 minutes, while other PRs' runs were being created normally. That was not a dropped event or a queue backlog. Master had moved (#1418, #1417, #1419) and the PR had gone to `mergeStateStatus: DIRTY`; GitHub cannot compute the `refs/pull/N/merge` ref for a conflicted PR, so `pull_request`-triggered workflows are never scheduled. The PR simply sits there looking like CI is asleep.

One real conflict, in `apps/electron-backend/src/app/events/stalker.events.ts`: #1418 added `requestUrlForLog` for compact credential-free logging in the same declaration block where the guard commit added `guardToken` / `countsTowardsGuard`. Both sides were wanted — kept both. `requestUrlForLog = fullUrl` still lands after the URL is built, so #1418's logging keeps its full URL and the guard keeps its host key. Everything else auto-merged.

Re-validated after the rebase: `shared-host-health` 37, `web-backend` 53, `web` 279, `electron-backend` 1674 (up 10, from #1418's own tests). Lint, `coverage:policy:check` and `release:notes:validate` all clean. `mergeStateStatus` is back to `UNSTABLE` and 14 checks are now running.

The CodeQL question from my previous comment is unchanged and still needs @4gray's call — I have not dismissed anything.

@greptileai

4gray:
Fixed in `48d6083e5` — thanks, this was a real bug and I'd missed the comment entirely on my first pass (I polled issue comments and Greptile, not the inline review threads).

**Propagate the discovery bypass through the PWA proxy** — confirmed end to end before fixing:

1. `StalkerPortalDiscoveryService.probeContent` sets `skipConnectionGuard: true` ✅
2. `PwaService.forwardStalkerRequest` did **not** forward it ❌
3. the web backend did **not** interpret it ❌

So on the PWA, two candidate probes hitting the 15 s timeout opened the breaker and the next candidate was fast-failed — exactly the failure mode the exemption exists to prevent, and worse than it sounds because discovery expects most candidates to fail and the one that works may be the last one tried. A working portal could be declared unreachable at import.

The flag now travels as a `/stalker` control param and the route applies all three Electron semantics: bypass the check, never count failures, **still report successes** so a candidate that answers clears what the others recorded. That last part is easy to drop and matters — `confirmFullPortal` runs the full non-exempt auth flow against auth-gated candidates. Like `macAddress`/`token`/`serialNumber`, the flag is stripped from the query forwarded to the portal.

Coverage: a backend test asserting two exempt probe failures do not open the breaker, that a later normal request still succeeds, and that `skipConnectionGuard` never appears in any portal-facing URL; plus two transport tests (forwarded when set, absent when not). Verified non-vacuous — forcing `countsTowardsGuard = true` fails exactly the exemption test. `web-backend` 54, `web` 281, lint/build/coverage-policy clean.

Also on this branch since the last review: the Windows E2E failure was `electron.launch: Electron failed to install correctly` — a runner-side Electron download failure, unrelated to this change (111 passed; macOS and Ubuntu green). Re-run triggered.

The CodeQL question is unchanged and still needs @4gray's call; nothing dismissed.

@greptileai
@codex review

4gray:
Fixed in `35fd626f2`. Both new findings were valid and turned out to share one root cause, so the fix is a scope reduction rather than two patches.

**Reset the guard when retrying an M3U URL import** (P2) and **Preserve the active half-open trial beyond 45 seconds** (P2) — both are consequences of my having extended the breaker to `/parse` and `/parse-xml`. Electron deliberately does not guard those: it wires the breaker into the two portal IPC handlers and nothing else. I checked before changing anything — `beginGuardedHostRequest` appears only in `xtream.events.ts` and `stalker.events.ts`; the playlist and EPG fetch paths have no guard at all.

Once that asymmetry is visible, both reports are the same bug:

1. `PLAYLIST_PARSE_BY_URL` is sent by the add-playlist dialog, so it is the direct result of the user asking. Refusing the retry locally for 30 s is a regression, not a protection — and unlike the portal Retry paths there is no natural reset site there.
2. A large XMLTV transfer legitimately runs for minutes, because the timeout is idle-based rather than a wall-clock deadline (documented in `host-guard.ts`). That outlives the 45 s trial expiry, which was sized against the 30 s portal ceiling, so a second trial could be admitted behind the first.

So the breaker now covers `/xtream` and `/stalker` only. Downloads keep the timeout, which is the fix those routes actually needed — they are one request, not a catalog fan-out, so the problem the breaker exists for does not arise. The reasoning is recorded under "Scope: portal calls only" in the contract doc, and the spec now pins the opposite of what it used to: four consecutive download failures, none of them fast-failed.

`web-backend` 54, `web` 281, lint/build/coverage-policy/release-notes clean.

---

**Distinguish same-millisecond sibling requests** (P2, from the first review) — valid, but I have deliberately not changed it, and I would like @4gray to decide.

The analysis is right: with `ECONNREFUSED` all failing inside one clock tick, sibling tokens can carry `startedAt === lastFailureAt`, the `>=` comparison counts them as consecutive, and one fan-out opens the breaker. A monotonic admission sequence instead of timestamps would fix it cleanly.

Two reasons I have left it alone. It is in the Electron commit (`cb63d8de9`), not in the PWA work this PR is about; and it is not an oversight there — the contract doc argues the tradeoff explicitly: *"Timestamps are millisecond coarse, so this only separates siblings once a request actually took time — which is precisely the expensive case worth protecting."* Rewriting the core streak logic to overturn another commit's documented decision, inside a PR about proxy timeouts, is not something I want to do silently.

Worth noting the doc's argument does not fully cover this case: it reasons about slow failures, where siblings are separated correctly, and does not address a fan-out of instant refusals imposing a 30 s window after a transient flap. So the finding stands — it just belongs either in the Electron commit's own review or in a follow-up.

---

Infrastructure, unrelated to the diff: **Build on linux x64 (portable)** failed with `RequestError: socket hang up` during electron-builder's download step, the same class as the earlier Windows `electron.launch: Electron failed to install correctly` (which passed on re-run). Re-running.

CodeQL remains as described — nothing dismissed, still needs your call.

@greptileai
@codex review

4gray:
Round 4. One of the three fixed in `2d0712f62`; the other two I have deliberately left, and they point at something structural worth deciding before merge.

**Admit the Xtream request before DNS revalidation** (P2) — valid, fixed. `normalizeAndValidateXtreamProviderUrl` resolves the hostname, and a dead host is exactly where DNS is slow or failing too, so an open breaker was paying for a lookup and could answer "host could not be resolved" instead of the fast-fail. I had originally ordered it that way out of caution that normalization might rewrite the host; checking `normalizeXtreamServerUrl` shows it rebuilds from `url.origin`, so the path can change but the host cannot. That caution was unfounded.

Admitting first introduces a leak the report does not mention: a request admitted and then abandoned by the URL policy would hold the half-open slot until the trial timeout expired. It now hands the token back through `releaseProviderRequest` — inconclusive, not a failure, which is what the contract already says policy refusals are. Two tests: an open host is fast-failed without any `resolveHostname` call, and a policy refusal during half-open does not consume the single trial. Verified non-vacuous.

`web-backend` 56, `web` 281, `shared-host-health` 37, lint and build clean.

---

**Keep slow portal trials active beyond 45 seconds** and **Reset the guard before Stalker catalog retries** — both valid, both left alone on purpose.

Neither is PWA work. They are in the Electron commit this PR carries (`cb63d8de9`), and they are the third and fourth findings against it, after **Distinguish same-millisecond sibling requests** and the redirect-attribution half I did fix. Taken together:

| Finding | Where |
| --- | --- |
| Redirect failures attributed to the wrong host | guard commit — fixed here, since the same bug was in my code |
| Same-millisecond siblings counted as a streak | guard commit — contradicts a documented tradeoff |
| Trial expiry by wall clock vs. idle-based timeouts | guard commit — `TRIAL_TIMEOUT_MS` assumes a 30 s wall-clock ceiling that axios does not provide |
| Stalker catalog append Retry never resets the host | guard commit — the breaker makes an existing Retry button inert for 30 s |

The last one is the most user-visible: the breaker turns a working Retry into one that silently does nothing for the rest of the window. The trial-expiry one is real on the portal routes too — a large Xtream catalog response can stream past 45 s without ever idling 30 s, so a second trial can be admitted behind the first.

Fixing them means changing the guard's core state machine and adding renderer reset sites, in a PR whose subject is proxy timeouts, and in one case overturning a tradeoff that commit's own doc argues for. That is a review that should happen against that commit, not underneath this one.

@4gray — `claude/trusting-leakey-ef53b9` passed CI on its own about an hour ago. Given it has four findings of its own, my recommendation is to open it as its own PR, land it first, and rebase this one onto it. That is also the cleanest answer to the duplicate-landing hazard in the PR description. If you would rather keep both commits here, say so and I will fix all four in this branch — I just do not want to rewrite another commit's documented design without you choosing that.

CodeQL is unchanged and still needs your call; nothing dismissed. Everything else is green — the three earlier red checks were all the same runner-side Electron install failure and cleared on re-run.

@greptileai
@codex review

4gray:
Rebuilt on master. `mergeable: MERGEABLE`, and the branch is now **two commits** instead of six.

**What happened:** #1421 merged, so the cherry-picked copy of the guard this PR was carrying became a stale duplicate of `e3f72f7dc` — hence the conflict. The duplicate-landing hazard from the PR description resolved itself the right way round.

Dropping the cherry-pick was not a plain rebase, because the guard did not merge as I had it. Its own review changed the shape underneath this work:

| Before (my copy) | On master now |
| --- | --- |
| `hostKeyOf` → `host:port` | `portalEndpointKeyOf` → origin, userinfo stripped |
| `token.host` | `token.endpoint` |
| redirect fix I added: compare hosts | `failedAfterRedirect` compares the **whole URL** |
| — | `reportGuardedHostFailure(token, error, { countFailures, requestUrl })` |

So I rebased by rebuilding rather than by resolving conflicts: reset to master, re-applied the PWA work, and redid the hoist against master's 624-line reviewed guard instead of my 479-line one. All six of my own files were untouched on master, so nothing of #1421's was lost — I checked each before restoring it.

**My redirect fix is gone, replaced by master's.** Master's is strictly better: comparing the whole URL catches a same-origin redirect (`/player_api.php` → `/slow/player_api.php`) that a host-only comparison misses. `host-guard.ts` now calls the shared `failedAfterRedirect`, and both proxy routes pass the URL they actually requested so a hop failure can be told from an endpoint failure.

**On the four findings I dispositioned back to you** — two were fixed upstream in #1421, which is where I said they belonged:

- ✅ Redirect attribution — fixed, better than my version.
- ✅ Stalker catalog append Retry — `retryContentPage()` now resets the guard before reloading (`with-stalker-content.feature.ts:875`).
- ❌ Same-millisecond siblings — still present (`token.startedAt >= state.lastFailureAt`).
- ❌ Half-open trial expiring by wall clock at 45 s — still present, and still contradicted by the idle-based timeout.

The last two are on master now, so they are no longer this PR's to carry. They are real, and worth a follow-up issue; they are not regressions introduced here.

**Validation after the rebuild:** `shared-host-health` 38, `web-backend` 56, `web` 281, `electron-backend` 1680 — plus lint, `web-backend` build, `coverage:policy:check` and `release:notes:validate` clean. One electron-backend run showed a single failure in `database worker zero-delay cancellation`; it passes in isolation and on a clean full re-run, and this branch does not touch the database worker.

The two commits are now `refactor: hoist the guard into libs/shared/host-health` and `fix: time out and fast-fail PWA proxy requests`, which should read far better than the six rounds of self-correction they replace.

CodeQL still needs your call — unchanged, nothing dismissed.

@greptileai
@codex review

4gray:
Thanks — the heads-up caught a real hole, and the structural warning was exactly right.

**Rebase:** done before this arrived, and by the method you describe — I dropped the cherry-pick and rebuilt rather than resolving line by line. I reset to master, re-applied only the PWA work, and redid the hoist against master's 624-line reviewed guard instead of my stale 479-line copy. All six of my own files were verified untouched on master before restoring them. `mergeable: MERGEABLE`, three commits.

**Fix 4 (redirect attribution):** my branch had its own weaker version — comparing hosts, which misses a same-origin redirect. Deleted; `host-guard.ts` now calls the shared `failedAfterRedirect`, and both proxy routes pass the URL they actually requested.

**Fix 5 (exempt probes must report):** I had the same hole, and it was worse here. Fixed in `7603bb880`. My exempt path skipped the report entirely rather than passing `countFailures: false`. Since this route sets **no `validateStatus`**, axios rejects *every* non-2xx with `error.response` attached — so a probe answered with 404 or 500 was dropped instead of clearing the record, and two counted failures either side of it read as consecutive and opened the breaker mid-discovery. `reportProviderRequestFailure` now takes `countFailures` to match your reporter, and both routes always report. Regression test asserts a probe answered with 500 keeps the endpoint reachable; reverting the fix fails exactly that test.

**Fix 6 — checked before duplicating, and it is *not* redundant.** Your classifier change covers `isStalkerProbeTimeout` and `isStalkerAuthFailureMessage`, but **`getStalkerRequestErrorStatus` still reads `.status` first and then regexes `HTTP Error (\d{3})`** — and that is precisely what the PWA path produces. `PwaService.forwardStalkerRequest` turns any HTTP-200 `{message, status}` envelope into `Error('HTTP Error 502: …')` *and* sets `.status = 502`, so a guard refusal would still read as "the endpoint answered" and send discovery walking every candidate. The PWA branch keyed on `isHostConnectivityFastFailMessage` stays, and its spec asserts the resulting error has no numeric status and no `HTTP Error <code>` in its message.

Worth noting for the shared contract: fix 6 protects the renderer against a *hostname* that matches an auth phrase, but the transport can still reintroduce a *status*. Those are two different ways to lose the classification, and only the first is fixed centrally.

**Fixes 1, 2, 3, 7:** inherited automatically — the hoist moves your file, so origin keying, `trialId`, the counted-failure threshold and the message format all come with it. Fixes 8–12 are renderer-side and untouched by this PR; I verified `retryContentPage()` resets before reloading.

**CI traps:** hit both. The `CONFLICTING` → no-checks-at-all one cost me two turns blaming an Actions queue before I checked `mergeStateStatus`. The Electron 43.3.0 flake appeared in three different disguises here — `Electron failed to install correctly` in unit specs and in Playwright `electron.launch`, and `RequestError: socket hang up` inside electron-builder — all cleared on rerun.

Two findings from this PR's review are still open on master, since they were never mine to fix: same-millisecond siblings (`token.startedAt >= state.lastFailureAt` still counts a fan-out as a streak) and the half-open trial expiring by wall clock at 45 s while the timeout is idle-based. Both are worth a follow-up issue.

@greptileai
@codex review

4gray:
Both valid, both fixed in `a87f502ea`. The first one is important beyond this PR.

**Read the final Axios request URL** — correct, and it means `failedAfterRedirect` has a transport assumption baked into it. I verified rather than reasoning from the axios source, because this logic had already flip-flopped once in this PR. A live server that 302s to a dead port, against the installed axios 1.19.0:

```
asked for          : http://127.0.0.1:63953/player_api.php
config.url         : http://127.0.0.1:63953/player_api.php   ← original
request._currentUrl: http://127.0.0.1:1/dead                 ← the hop that failed
```

`config.url` is right for the Electron transport, which sets `maxRedirects: 0` and reissues each hop as its own request. It is blind on the web backend's default transport, where follow-redirects walks the chain inside one request and `config` is built once. So the comparison was original-vs-original, found no redirect, and charged two dead destinations to the provider that had answered both times with a 302 — then fast-failed it. Exactly the failure the helper exists to prevent, inverted by the transport.

Fixed in the shared helper: read `request._currentUrl` first, fall back to `config.url`. Both transports are covered, and Electron's native per-hop requests expose no `_currentUrl`, so its behaviour is unchanged — `electron-backend` is still 1680 green.

Worth flagging to @4gray and the #1421 author: **this was latent in master too.** Nothing on master uses a following transport today, so it is not a live bug there, but the helper silently depended on `maxRedirects: 0` without saying so. It now says so, and works either way.

It also caught a weak test of mine: my web-backend redirect spec was using the `config.url` shape, i.e. the Electron transport's, so it passed through the fallback path rather than the one this route actually produces. Now uses the real shape.

**Credit redirects observed by exempt probes** — also correct. `if (!countFailures) break;` sat before the redirect check, so a probe redirected to a dead destination never credited the guarded endpoint's 3xx. Reordered: attribution first, suppression second. Note the same ordering is in master's `reportGuardedHostFailure`; I have not touched it there, for the same reason as before — it is merged, reviewed code and not this PR's subject — but it is the identical bug and worth a follow-up.

Both fixes verified non-vacuous: disabling the `_currentUrl` read fails exactly the two redirect tests, and the earlier `countFailures` revert fails exactly the exempt-probe test.

`shared-host-health` 41, `web-backend` 57, `web` 281, `electron-backend` 1680, lint clean.

CI on the previous head finished 18 green with CodeQL the only failure; those two alerts are unchanged and still need your call.

@greptileai
@codex review

4gray:
Correct, and this was the most serious finding on the PR so far — the Xtream breaker was not merely mis-attributing, it **never opened at all**. Fixed in `2f67a1e0e`.

Verified against axios 1.19.0 with a plain `ECONNREFUSED` and no redirect anywhere in sight:

```
baseline (what the route built) : http://127.0.0.1:1/player_api.php
request._currentUrl (what was sent): http://127.0.0.1:1/player_api.php?username=demo&action=get_account_info
```

The route passes credentials and the action through axios' `params`, so the sent URL always carries a query the baseline does not. Two different normalized URLs, every ordinary failure classified as post-redirect, endpoint credited, breaker permanently disarmed. My previous commit fixed *which field* to read and in doing so activated this — reading `config.url` had made the comparison a no-op, so the bug only became reachable once the comparison started working.

**Fix:** compare origin and path, not the whole URL. That keeps what the check is for — an endpoint that answered and sent us somewhere else, including the same-origin `/player_api.php` → `/slow/player_api.php` case the helper was written for — and gives up only a redirect that changes nothing but the query, which is then counted as an ordinary failure. Erring towards counting is the safe direction; erring the other way is what produced this.

**The part worth keeping from this round is why 57 tests passed over a dead feature.** `StubHttpClient` threw bare `Error`s, so the guard's redirect check saw neither `config.url` nor `request._currentUrl` and quietly did nothing — the tests were exercising a code path that cannot occur in production. The stub now shapes its rejections like axios does, including the query axios appends. With that change alone and the old comparison restored, **four existing tests fail**:

- fast-fails Xtream requests … once the host stops answering
- fast-fails an open host before spending a DNS lookup on it
- lets exactly one request through once the open window elapses
- contacts the host for real again after an explicit reset

So the coverage was real; the fixture was not. That is the more durable fix here.

@4gray — worth noting for the merged guard: `failedAfterRedirect` on master compares whole URLs too. It is not a live bug there, because both Electron call sites build their query into the URL before handing it to axios and pass that same string as the baseline. But it holds only as long as no caller uses axios `params`, and nothing says so. The shared helper is now immune either way.

`shared-host-health` 41, `web-backend` 57, `web` 281, `electron-backend` 1680, lint clean. Greptile was 5/5 on `a87f502ea`; this commit needs a fresh pass.

@greptileai
@codex review

4gray:
Both fixed in `f6d9ead1d`.

**Count DNS revalidation failures against the host** (P2) — valid, and it was a hole I opened myself two commits ago. A name that will not resolve is the host failing to answer, the same evidence as the `ENOTFOUND` the transport would have raised a moment later; but the SSRF validation converts a lookup failure into a 400, and the release path I added handed the token back as inconclusive. So the breaker could never open for a host whose DNS died, and every request kept paying for the same dead lookup.

`ProviderUrlError` now carries the underlying lookup error internally. A refusal that has one is counted; a genuine policy refusal — private address, bad scheme, URL credentials — still only releases the half-open slot, because that genuinely says nothing about reachability. The distinction matters: counting an SSRF refusal would let a misconfigured target fast-fail a healthy host.

The field is internal. `providerUrlErrorBody()` strips it at both call sites — the `/xtream` route and `/provider-targets` — so the client sees exactly the body it saw before. The test asserts the response is still precisely `{ message: 'Provider URL host could not be resolved', status: 400 }`, so the internal channel cannot leak into the wire contract.

**Update the shared reset helper's PWA contract** (P3) — valid. The docblock still said the PWA channel is unknown and the call no-ops, which stopped being true when this branch implemented `CONNECTIVITY_GUARD_RESET` over HTTP. Worth recording how it came back: I had fixed it in an earlier commit, and the edit was lost when the branch was rebuilt on master after #1421 merged. Restored, and reworded so the next caller is told there is no longer a runtime where the helper quietly does nothing.

Verified non-vacuous: replacing the branch with an unconditional release fails exactly the new DNS test.

`shared-host-health` 41, `web-backend` 58, `web` 281, `services` 368, `electron-backend` 1680; lint and build clean.

---

Where this stands, plainly. Greptile has been 5/5 on each head it has actually reviewed, including `2f67a1e0e`. Codex has found a real defect every round, but the severity is falling — P1 (breaker never opened), then P2 + P3 this round, and the P3 is documentation. The two open items are both yours rather than mine: the CodeQL pair, unchanged and undismissed, and the two findings that now live on master (same-millisecond siblings, and the 45 s wall-clock trial expiry against an idle-based timeout).

@greptileai
@codex review

4gray:
Fixed in `333875431`, and this one was worth taking seriously despite being P3 — a canonical contract that describes one transport while two are in play is how the attribution bugs from the last two commits get reintroduced.

The section had three claims, two of which stopped being true for the web backend. It now states the split explicitly:

| | Electron | Web backend |
| --- | --- | --- |
| Redirects | followed hop by hop (`maxRedirects: 0`), each its own request | followed inside one request by follow-redirects |
| Failed hop is in | `error.config.url` | `error.request._currentUrl` |

…plus why `failedRequestUrlOf` reads `_currentUrl` first and falls back to `config.url` (a per-hop request exposes no `_currentUrl`; on the following transport `config` is built once and keeps the URL we asked for), and why the comparison is origin + path rather than the whole URL (credentials ride in axios' `params`, so the sent URL always carries a query the baseline does not).

---

That closes every actionable bot finding on this PR. Summary of the review, since it ran long:

| Round | Finding | Outcome |
| --- | --- | --- |
| 1 | Redirect failures poison source host (Greptile P1) | fixed, then superseded by the version that landed in #1421 |
| 2 | Discovery bypass dropped in the PWA (P2) | fixed |
| 3 | `/parse` retry refusal + 45 s trial vs idle timeout (P2 ×2) | fixed by scoping the breaker to the portal routes |
| 4 | Admit before DNS revalidation (P2) | fixed; also fixed the slot leak it introduced |
| 5 | Exempt probes must report (from #1421's author) | fixed |
| 6 | Read the final axios URL (P2) + credit redirects for exempt probes (P2) | fixed |
| 7 | **Axios params read as a redirect (P1) — breaker never opened** | fixed, plus the unrealistic stub that hid it |
| 8 | DNS revalidation not counted (P2), stale reset docblock (P3) | fixed |
| 9 | This doc contract (P3) | fixed |

Three of those were bugs an earlier fix of mine made reachable, which is worth saying out loud: reading `config.url` masked the params bug until the transport fix activated it, and the release path added in round 4 created the DNS gap closed in round 8.

The most useful outcome is not a code change: `StubHttpClient` was throwing bare `Error`s, so the guard's redirect logic saw no `config.url` and no `_currentUrl` and silently did nothing. 57 tests passed over a breaker that could not open. The stub now shapes rejections like axios, including the query axios appends — and with that alone, four existing tests fail against the old comparison. The coverage was real; the fixture was not.

**Still open, and both yours:**

1. The two CodeQL alerts — pre-existing, unchanged, nothing dismissed. My recommendation stands: dismiss as pre-existing and close the redirect-hop validation gap in its own PR, where `beforeRedirect` is not sufficient because DNS rebinding walks through it.
2. Two findings that now live on master, not here: same-millisecond siblings counting a fan-out as a streak, and the half-open trial expiring by wall clock at 45 s while the timeout is idle-based. Worth a follow-up issue against the merged guard.

@greptileai
@codex review

4gray:
Fixed in the latest commit — the opening line still called the guard an Electron main-process concern, which contradicted the ownership section below it and is exactly the line a reader skims to decide whether the document applies to them. It now names both processes, and records that the web backend had the worse version of the problem first (no request timeout at all), since that is why the timeouts and the breaker had to land there together.

I also checked the rest of the document for the same Electron-only framing rather than fixing only the line that was flagged; the summary was the last of it.

That is the second documentation-only round in a row, with no behavioural finding since `2f67a1e0e`. Both were worth doing — this is the canonical contract, and the redirect section in particular is what stops the attribution bugs from round 6 and 7 being reintroduced — but the loop has converged.

**Final state:** 19 CI checks green with CodeQL the only failure; Greptile 5/5 on every head it has reviewed. Locally: `shared-host-health` 41, `web-backend` 58, `web` 281, `services` 368, `electron-backend` 1680, plus lint, build, coverage policy and release notes.

**Two decisions remain, and both are @4gray's:**

1. **The two CodeQL alerts.** Pre-existing and undismissed. All four proxy request sites in `web-backend-app.ts` carry the same suppression comment and the same behaviour; CodeQL is green on master with identical code, and the two reported here are the two whose call line this PR touched. The real gap behind them — the web backend not re-validating redirect hops — is worth its own PR, where `beforeRedirect` is not a sufficient fix because DNS rebinding walks straight through it.
2. **Two findings that now live on master**, surfaced by this review but belonging to the merged guard: same-millisecond siblings counting one fan-out as a streak, and the half-open trial expiring by wall clock at 45 s while the timeout is idle-based. Neither is a regression from this PR.

Nothing else is outstanding from my side.

4gray:
CodeQL alerts dismissed and the follow-up filed as #1436.

**Dismissed:** #74 (`/xtream`), #75 (`/parse`), #76 (`/stalker`) — a third appeared during the review as more `.get()` call lines were touched, which is itself evidence for the diagnosis. Reason recorded as **won't fix**, not *false positive*: the finding is not wrong. All four proxy sites carry the same pattern, CodeQL is green on master with identical code, and these surfaced here only because the PR check reports alerts located on changed lines.

`CodeQL — No new alerts in code changed by this pull request`. The PR is now `MERGEABLE / CLEAN`.

**#1436** covers the real gap: the web backend validates a provider URL once at `/provider-targets` and then lets axios follow up to 21 redirects without re-validating any of them, so a registered public URL can `302` to a link-local address. It records why `beforeRedirect` is not an acceptable fix — synchronous checks stop the cloud-metadata case but DNS rebinding walks straight through, and shipping that as "fixed" is worse than a documented gap — and sketches the shape that would work: a wrapper implementing the existing `WebBackendHttpClient` interface that mirrors `validated-axios.ts`, with `resolveHostname` injected from `WebBackendAppOptions`. It also flags that a redirect-following wrapper will need `StubHttpClient` to model 3xx, and why fixture realism matters here specifically.

I have also rewritten the PR description, which still opened with a warning that this branch carried a cherry-pick of #1421 and had to be merged in a particular order. That stopped being true when #1421 merged and the branch was rebuilt on master; leaving it there would have told a reviewer to do something already done.

**Two things remain, neither of them mine to close:**

1. `AGENTS.md` / `CLAUDE.md` say final summaries should state test impact — done throughout, and the release note is `.changes/portals-pwa-request-timeouts.md` (`type: fix`, since this is a user-visible hang rather than only a perf win).
2. Two findings raised during this review live on master, not here: same-millisecond siblings counting one fan-out as a streak, and the half-open trial expiring by wall clock at 45 s while the timeout is idle-based. Neither is a regression from this PR, and neither has an issue yet — say the word and I will file them against the merged guard.

Nothing else outstanding from my side. I have not merged.

4gray:
Both remaining findings are now filed against the merged guard: #1438 (one fan-out of instant failures can trip the breaker) and #1439 (half-open lets a second trial out when one legitimately runs past 45 s).

Worth noting on #1439: I scoped it down from the original report after checking the shipped code. The state-corruption half — an abandoned trial freeing or clobbering the replacement's slot — was **already fixed** by the `trialId` work in #1421, and `ownsTrial` gates it. What is left is only that a second request goes out during half-open against a slow endpoint. Lower severity than reported, so the issue says so rather than repeating the original framing.

Both issues cite the specific reason the premise fails: `requestWithValidatedRedirects` gives each of up to five hops its own 30 s budget, and axios' timeout is idle-based, so neither has a 45 s ceiling. The `trialId` docblock in that file already says the first half out loud — it was never reconciled with the constant a few lines above it.

Together with #1436 (redirect-hop re-validation, from the dismissed CodeQL alerts), that is everything this review surfaced which does not belong to this PR. Nothing outstanding from my side; #1424 is `MERGEABLE / CLEAN` with all checks green, and I have not merged.

4gray:
Rebased onto current master (`d90adf59f`) — it had moved again (#1420, #1426, #1431, #1432, #1433) and the PR went `CONFLICTING` while the issues were being filed.

One conflict, in `CLAUDE.md`, and it was a genuine two-sided one rather than a formatting clash: #1431 rewrote the reset-senders list to name the extracted `XtreamRefreshFlowService` (replacing the two duplicated refresh implementations), while this branch had edited the same sentence to say the breaker's rules now live in `@iptvnator/shared/host-health` and that the web backend runs them too. Taking either side wholesale would have silently dropped the other's fact, so I kept master's line and applied only my delta onto it. Verified afterwards that all three survive: `XtreamRefreshFlowService`, the shared-lib location, and the web-backend clause.

Worth noting that #1431 is the deduplication follow-up that came out of #1421's review — the two ~60-line copies of the destructive Xtream refresh are now one flow, so the "a rule applied to one of two near-identical flows LOOKS applied" trap is closed at the source.

Re-validated against the new master: `shared-host-health` 41, `web-backend` 58, `web` 284, `services` 368, `electron-backend` 1680; lint, coverage policy and release notes clean. `MERGEABLE` again, CI running.

4gray:
All 19 checks are green on `d90adf59f`, including CodeQL (now reporting no new alerts after the pre-existing ones were dismissed) and Electron E2E on all three platforms.

Retriggering review: the rebase onto current master rewrote the commit SHAs, so the last Greptile pass sits on `f6d9ead1d`, which no longer exists in the branch. The code is unchanged by that rebase apart from the two-sided `CLAUDE.md` merge, but a 5/5 against a SHA that is not the head is not a current verdict, so it is worth a fresh pass rather than being quoted as one.

@greptileai
@codex review

4gray:
Valid, fixed in `c89596b7a` — and my check of the review had missed it: Codex posted this at 17:02:32Z, about a minute after my previous poll window closed, so my last summary claimed the review was clean when it was not. Re-checked properly this time.

**Declare the shared interfaces runtime dependency** (P2). The new library's manifest listed only `tslib`, while its emitted JavaScript does `require('@iptvnator/shared/interfaces')` — the guard builds its refusal text with `buildHostConnectivityFastFailMessage`. Anything resolving that artifact from its own manifest would fail with MODULE_NOT_FOUND.

How it happened is worth recording: I copied the manifest from `shared/logging`, which imports nothing across libraries and so legitimately needs nothing beyond `tslib`. That made it a bad template for a library that does. `shared/m3u-utils` is the correct precedent — it imports the same library and declares `"@iptvnator/shared/interfaces": "0.0.1"`, which is exactly what has been added.

Verified against the build output rather than by reading the manifest: `dist/libs/shared/host-health/src/lib/host-connectivity-guard.js` does require the module, and the generated `dist/.../package.json` now declares it.

`shared-host-health` 41, `web-backend` 58, lint and both builds clean.

Master has moved three more commits since the rebase (#1423, #1435, #1437) but the branch still reports `MERGEABLE / CLEAN`, so no rebase is needed for those.

@greptileai
@codex review

4gray:
Valid, fixed in `f04218c04`. This one is worth naming plainly: **it is the bug this PR exists to fix, reintroduced one layer up by the fix itself.**

The reset was a bare `fetch` with no timeout. Every caller awaits it *before* issuing the request it is clearing the way for — `retryContentInitialization` awaits it first by design, because a reset placed any later never runs — so a backend or reverse proxy that accepts the POST and then goes quiet would leave Retry doing nothing at all, for as long as the socket stayed open. The helper's own docblock promises that a failed reset must never block the action that asked for it; an unbounded `fetch` cannot keep that promise, because it never fails.

Bounded with an `AbortController` and a 5 s timer. The abort rejects, `resetHostConnectivityGuard` swallows it exactly as it already does for any other failure, and the caller proceeds to its real request. The timer is cleared in a `finally` and covers the body read as well as the headers. Five seconds because this talks to the user's own backend rather than a provider — it should answer immediately, and a slow one must not hold up the retry.

Regression test drives it with fake timers: a `fetch` that only ever settles via abort, advanced past the timeout, asserting the signal was passed, that it starts unaborted, and that the promise rejects with `AbortError`. Removing `signal: controller.signal` fails exactly that test.

`web` 285, `web-backend` 58, `shared-host-health` 41; lint and the web build clean.

Greptile is 5/5 on `c89596b7a`, so this commit needs a fresh pass.

Two of the last three findings have been the same shape — a template or habit carried over without re-checking whether it fitted: `shared/logging`'s manifest for a library that does import across libraries, and a bare `fetch` in a change about unbounded requests. Both passed every test.

@greptileai
@codex review

PR #1425, opened 2026-08-12
fix(dashboard): keep every catalog row per title key when matching

4gray:
## The bug

A catalog holding two cuts of one film — "Dune 1984" and "Dune 2021" — produces two rows under the same normalized title key. `buildTitleMatchIndex` collapsed them to **one row before the caller's year was known**, keeping whichever the worker returned first. The caller's `titleYearsCompatible` check then rejected the retained row on behalf of the copy the user actually owns, and the card rendered as *not in your library* for a movie the library holds.

The year that separates same-titled rows belongs to the **lookup**, which the grouping cannot see — so the grouping must not decide.

## The fix

`buildTitleMatchIndex` is replaced by a shared pair in `libs/services/src/lib/catalog-title-match.service.ts`:

- `groupTitleMatchesByKey()` — keeps **every** row per `type:exactNormalizedTitle`
- `pickTitleMatch({type, titles, year}, grouped)` — ranks the year-compatible rows by evidence: exact year → untagged → anything else compatible

`titles` takes aliases most-trusted-first and ranks across all of them at once, so a weak hit under one alias cannot veto a strong one under another.

## Scope: this was four bugs, not one

The reported symptom was the Trending rail, but the same `index.get(key)` → `titleYearsCompatible` shape was copy-pasted into every consumer, each with the identical latent bug:

| Consumer | Symptom |
| --- | --- |
| `DashboardTrendingService` | card renders unmatched, links to global search |
| `CrossPortalSimilarService` | recommendation silently dropped from the "Similar" rail |
| `XtreamActorRouteComponent` | filmography row loses its "In your library" badge |
| `StalkerActorRouteComponent` | same |

All four now share the helpers. With no consumers left, the collapsing helper is deleted rather than left exported from `@iptvnator/services` to invite a fifth instance.

**Second commit** folds in the "Because you watched" rail (#1419), which merged to master while this was in progress. It had shipped its own `groupMatchesByKey`/`pickCatalogMatch` written specifically to dodge the collapsing index; those are behaviourally identical to the shared pair, so the rail now delegates through a small `candidateLookup()` adapter. `pickTitleMatch`'s alias array was designed for that caller — it is the only one with a second title to pass (TMDB `original_title`).

## Tests

New `catalog-title-match.service.spec.ts` for the helper, plus rail-level regressions in the trending and cross-portal specs.

Verified by **mutation**, four of them, not one:

| Mutation | Caught by |
| --- | --- |
| restore the collapse | 5 tests, incl. both rail-level regressions |
| drop the untagged tier | 2 tests |
| drop the year filter | 3 tests |
| reverse alias order | 1 test |

Restoring the collapse *after* the second commit also fails the recommendations rail's own two year-ranking tests — confirming the delegation is real and not accidental duplication.

Two tests I wrote survived **all four** mutations, meaning they could not fail under any mutation of the logic they named. One was deleted; the other had its fixture reworked until it genuinely pins the tier ordering. (Same failure class flagged during #1419's review.)

## Validation

- `pnpm nx test workspace-dashboard-data-access` — 99/99
- `pnpm nx test services` — 359/359
- `pnpm nx build web` — clean (the `lib: es2018` gate; no `flatMap`)
- `pnpm nx affected -t lint` — 28 projects, 0 errors
- `pnpm run release:notes:validate` — 104 valid

## Docs

Release note added (`type: fix`). New "Resolving a batched match" section in `docs/architecture/tmdb-metadata-enrichment.md` and a matching CLAUDE.md line, both stating why the grouping must not collapse — plus a corrected comment in `title-sources.operations.ts` that justified multi-source discovery against the deleted helper.


4gray:
Context from #1419, which is now merged (`4bcd4bd3`) — so the delegation target this PR folds onto is on master, and `3803b7f0c` does exactly what was left open there.

I re-read the folded code against what #1419's private matcher did, since that behaviour took 14 review rounds to settle and a faithful-looking refactor is the easiest place to lose it. It holds, and here is where each property now lives, so a later change does not undo one by accident:

- **Year-compatibility filters BEFORE ranking**, not after — `pickTitleMatch` pushes only `titleYearsCompatible` rows into the pool. Filtering after would let an incompatible row win and drop the card.
- **Ranking is across all aliases at once**, not per alias: one pool, `year-exact → untagged → any compatible`. A bad hit under the localized title must not veto a good one under the original-title alias (#1419 round 14).
- **Alias order survives only as the tiebreaker** inside a tier — rows enter in `lookup.titles` order, so `find` still prefers the localized title where evidence is equal.
- **Grouping keeps every row per key** rather than collapsing to one, which is the actual bug this PR fixes for the other three consumers (#1419 round 6).
- **`candidateLookup` passes both aliases and the candidate's year**, which is what `titles[]` exists for.
- **Row dedupe stayed in the rail** (`claimedRows` in `attachMatches`), not in the matcher — correct: same-titled remakes must both reach the matcher and only collapse if they resolve to the same catalog row (#1419 round 8).

The alias-key dedupe inside `pickTitleMatch` is new and a genuine improvement — TMDB repeats the title as `original_title` for English-language items, so the second pass was always wasted there.

One process note, since it bit both PRs: mutation-verify each regression test by reverting the fix and confirming the test fails. In #1419 that caught two of my own tests that passed with the fix reverted — one asserted `not.toContain` on a list that was empty for an unrelated reason, the other ordered its fixture so the good candidate survived the old code anyway. A test that never fails under any mutation of the code it names is not testing that code.

4gray:
@codex review

PR #1427, opened 2026-08-12
ci(embedded-mpv): add pinned mirrors to the Linux runtime source download

4gray:
## Why

The **Build pinned Linux Embedded MPV runtime** job failed twice on 2026-08-11 (runs 31541059270 and 31546729212, during #1417) because `https://www.freedesktop.org/software/fontconfig/release/fontconfig-2.16.0.tar.xz` answered the GitHub runner with **HTTP 418**. A rerun passed, so this is upstream rate-limiting rather than a broken pin.

`downloadArchive()` in `build-linux-runtime.mjs` curled a single `sourceUrl` with no fallback. The macOS builder already solved this with the shared `downloadPinnedSource()` helper, which walks a URL list and validates the pinned SHA-256 for each candidate.

## What changed

- **`build-linux-runtime.mjs`** — `downloadArchive()` now goes through `downloadPinnedSource({ archivePath, expectedSha256, urls, download })`. The curl hardening flags (`--proto =https`, `--tlsv1.2`, `--fail`, `--location`) move into the download callback unchanged, and `assertArchiveMatchesPin` still runs before extraction. A mirror that actually served the bytes is reported in the build log.
- **`build-linux-runtime.cjs`** — `mirrors` for the three single-host pins. Each candidate was downloaded and confirmed to hash to the existing pin:

  | package | mirror | pin |
  |---|---|---|
  | fontconfig 2.16.0 | `ftp.osuosl.org` (BLFS conglomeration) | `6a33dc55…` ✓ |
  | libdisplay-info 0.1.1 | `ftp.osuosl.org` (BLFS conglomeration) | `0d873158…` ✓ |
  | freetype 2.13.3 | SourceForge | `05503506…` ✓ |

  The remaining archive pins are on GitHub / ffmpeg.org. freetype is the one package outside the freedesktop scope — savannah is the same kind of single point of failure and the macOS builder already pins this exact SourceForge URL for the same tarball.

## Supply chain

No weakening. The Linux side already pinned and verified checksums via `assertArchiveMatchesPin`, and the shared helper verifies **every** candidate against the same pin, discarding a mismatch instead of using it. A mirror serving different bytes is rejected.

## One deliberate divergence from macOS

macOS records the *resolved* URL in its manifest (`sourcePackage.sourceUrl = sourceUrl`). Linux must not: `validateRuntimeManifestPackages` requires the manifest `sourceUrl` to deep-equal the immutable pin, and the Snap publication boundary cross-checks that identity across manifest / license-inputs / notices. Recording a mirror there would break notice generation and Snap publication on any run that fell back. The Linux record keeps the canonical URL and `mirrors` never reaches the manifest — both locked down by tests.

## Release-chain follow-through

`build-linux-runtime.mjs` now *imports* the downloader, so `download-pinned-source.mjs` was added to:

- `SOURCE_TOOLING_FILES` in `release-snap-source-binding.cjs`, its workflow `cp` list, and the `release-snap-assets.test.mjs` fixture — otherwise `linux-frame-copy-runtime-sources.tar.xz` would ship a build script it cannot run, against the LGPL "complete corresponding build scripts" claim;
- the Linux runtime cache key's `hashFiles(...)` — otherwise a downloader change would not invalidate a cached runtime.

## Tests

**`build-linux-runtime.test.mjs`** (behavioral, driving the now-exported `downloadArchive` with a fake `context.run`):
- mirror fallback — attempt order, curl hardening flags per attempt, canonical `sourceUrl` in the record, archive written, exactly one extraction;
- pin mismatch on every URL — throws, nothing extracted, no archive left behind;
- the existing manifest test now asserts no package leaks `mirrors` and every `sourceUrl` stays canonical.

**`download-pinned-source.test.mjs`**:
- the Linux routing contract (incl. that the resolved mirror never overwrites the recorded URL);
- mirror hygiene — https, different host, same basename, and **every** freedesktop-hosted pin must carry a mirror;
- workflow ↔ released-tooling consistency, so the two lists cannot drift silently.

Each new assertion was mutation-tested: removing the fallback, recording the mirror URL, leaking `mirrors` into the manifest, dropping the file from the `cp` list, and dropping a mirror each fail the intended test. The first draft of the tooling test survived mutation (it only proved the filename appeared *somewhere* in the workflow) and was tightened to parse the actual `cp` block.

## Validation

- `pnpm nx test packaging --skip-nx-cache` → **261/261 pass**
- `pnpm nx lint packaging` → clean; prettier clean on all changed files
- `pnpm nx test release-tools` → pass
- `pnpm nx test electron-backend` → 1651/1652; the single failure is `worker-performance-capture.concurrency.spec.ts`, the known flaky event-loop-delay spec, which passes on isolated re-run and touches nothing here
- `electron-backend-e2e` skipped: no app runtime code changed, affected only via the workflow-file input

## Docs / release note

`tools/embedded-mpv/README.md` gained the Linux acquisition + mirror contract next to the existing macOS paragraph. No `.changes/` note: the release-note gate triggers only on `apps/` or `libs/`, and this is CI plumbing with no user-visible behavior.

### DIFF
--- a/.github/workflows/build-and-make.yaml
+++ b/.github/workflows/build-and-make.yaml
@@ -80,7 +80,7 @@ jobs:
                       echo 'meson=1.7.2'
                   } > "${RUNNER_TEMP}/linux-runtime-toolchain.txt"
                   TOOLCHAIN_SHA256="$(sha256sum "${RUNNER_TEMP}/linux-runtime-toolchain.txt" | cut -d ' ' -f 1)"
-                  SOURCE_SHA256="${{ hashFiles('tools/embedded-mpv/build-linux-runtime.cjs', 'tools/embedded-mpv/build-linux-runtime.mjs', 'tools/embedded-mpv/generate-linux-runtime-notices.cjs', 'tools/embedded-mpv/linux-runtime-manifest.cjs', 'tools/embedded-mpv/linux-source-archive-contract.cjs', 'tools/embedded-mpv/stage-runtime.mjs', 'tools/packaging/prepare-linux-runtime-source-snapshot.cjs') }}"
+                  SOURCE_SHA256="${{ hashFiles('tools/embedded-mpv/build-linux-runtime.cjs', 'tools/embedded-mpv/build-linux-runtime.mjs', 'tools/embedded-mpv/download-pinned-source.mjs', 'tools/embedded-mpv/generate-linux-runtime-notices.cjs', 'tools/embedded-mpv/linux-runtime-manifest.cjs', 'tools/embedded-mpv/linux-source-archive-contract.cjs', 'tools/embedded-mpv/stage-runtime.mjs', 'tools/packaging/prepare-linux-runtime-source-snapshot.cjs') }}"
                   echo "toolchain-sha256=${TOOLCHAIN_SHA256}" >> "${GITHUB_OUTPUT}"
                   echo "key=linux-frame-copy-runtime-v5-ubuntu-22.04-${TOOLCHAIN_SHA256}-${SOURCE_SHA256}" >> "${GITHUB_OUTPUT}"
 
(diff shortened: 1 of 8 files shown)


4gray:
@codex review

PR #1430, opened 2026-08-13
refactor(portals): extract the destructive Xtream refresh into one flow

4gray:
> Stacked on #1421 — base is `claude/trusting-leakey-ef53b9`, not `master`. Review only the top commit; merge after #1421.

## Why

#1421 had to fix the same bug twice. `PlaylistRefreshActionService.refreshXtream()` and `RecentPlaylistsComponent.refreshXtreamPlaylist()` were two independent ~60-line implementations of the same destructive refresh, so the connectivity-guard reset added to the first one *looked* applied to both — the sources page kept deleting the cached catalog and then bootstrapping a re-import that an open guard fast-failed, leaving the user with nothing until the cooldown expired.

The guard contract even documented the duplication as a standing hazard ("There are **two independent implementations** of this flow and both need the reset"). This removes the hazard instead of documenting it.

## What

`XtreamRefreshFlowService` (new) owns the sequence once: confirm → reset the guard → delete the catalog while collecting playback positions and stamping the update date → park the restore state → dispatch the meta update → navigate to re-import → error/abort handling.

The entry points differ only in how they report progress, so that is injected as an `XtreamRefreshProgressReporter` (`isBusy` / `begin` / `report` / `end`, plus an optional `waitForVisibleProgress`):

- **`PlaylistRefreshActionService`** keeps its global `refreshPreparation` signal, its operation-scoped event filter and its paint delay (through the optional hook).
- **`RecentPlaylistsComponent`** keeps its per-row `pendingRefreshIds` / `busyOperations` indicators, which it shares with deletion.

A reporter cannot reach the guard reset — it is the first awaited statement in the shared flow — so a third entry point gets it for free. The action service also drops 393 → 309 effective lines, off the edge of the 400-line cap.

## Behavior delta

One, deliberate: the sources page now awaits `router.navigate` before clearing its busy row, matching the action service. It previously cleared the row while navigation was still in flight. Not user-visible, but it is not a pure extraction.

## Tests

- New `xtream-refresh-flow.service.spec.ts` pins the reporter contract: re-entry after confirmation, busy-before-reset ordering, progress forwarding under one run, abort vs error, and a reporter that implements no paint hook.
- Both existing guard-reset regression tests were kept meaningful, not just green: removing the reset from the shared flow fails **exactly two** tests, one per entry point. Mutation-checked the new spec the same way (dropping `isBusy`, moving `begin()` after the reset).
- The sources spec's abort test counted a fixed number of microtasks to reach the delete; the shared flow adds one await to that path, so it now waits on a deterministic signal, matching the sibling spec's convention.

| Check | Result |
| --- | --- |
| `pnpm nx test playlist-shared-ui` | 98 passed (91 existing + 7 new) |
| `pnpm nx lint playlist-shared-ui` | clean |
| `nx test workspace-shell-feature` / `workspace-dashboard-feature` | 51 + 144 passed |
| `nx build web` | passed |
| `electron-backend-e2e:e2e-ci--src/sources.e2e.ts` | 4 passed — covers this refresh in the real app |

## Docs

Updated both places that recorded the duplication as a fact: the "Explicit reset" call-site list in `docs/architecture/host-connectivity-guard.md` and the matching sentence in `CLAUDE.md`. Both now describe one flow with two entry points, and say a third should pass a reporter rather than copy the sequence.

No `.changes/` note — pure refactor, hence the `no-release-note` label.


4gray:
Superseded by #1431 — this PR was auto-closed when its base branch `claude/trusting-leakey-ef53b9` was deleted on the #1421 merge. Same single commit, rebased onto `master`.

PR #1431, opened 2026-08-13
refactor(portals): extract the destructive Xtream refresh into one flow

4gray:
Follow-up to #1421, which merged while this was in review — #1430 was auto-closed when its base branch was deleted, so this is the same single commit rebased onto `master`. The touched paths were byte-identical between #1421's tip and the squash on master, so the rebase was clean.

## Why

#1421 had to fix the same bug twice. `PlaylistRefreshActionService.refreshXtream()` and `RecentPlaylistsComponent.refreshXtreamPlaylist()` were two independent ~60-line implementations of the same destructive refresh, so the connectivity-guard reset added to the first one *looked* applied to both — the sources page kept deleting the cached catalog and then bootstrapping a re-import that an open guard fast-failed, leaving the user with nothing until the cooldown expired.

The guard contract even documented the duplication as a standing hazard ("There are **two independent implementations** of this flow and both need the reset"). This removes the hazard instead of documenting it.

## What

`XtreamRefreshFlowService` (new) owns the sequence once: confirm → reset the guard → delete the catalog while collecting playback positions and stamping the update date → park the restore state → dispatch the meta update → navigate to re-import → error/abort handling.

The entry points differ only in how they report progress, so that is injected as an `XtreamRefreshProgressReporter` (`isBusy` / `begin` / `report` / `end`, plus an optional `waitForVisibleProgress`):

- **`PlaylistRefreshActionService`** keeps its global `refreshPreparation` signal, its operation-scoped event filter and its paint delay (through the optional hook).
- **`RecentPlaylistsComponent`** keeps its per-row `pendingRefreshIds` / `busyOperations` indicators, which it shares with deletion.

A reporter cannot reach the guard reset — it is the first awaited statement in the shared flow — so a third entry point gets it for free. The action service also drops 393 → 309 effective lines, off the edge of the 400-line cap.

## Behavior delta

One, deliberate: the sources page now awaits `router.navigate` before clearing its busy row, matching the action service. It previously cleared the row while navigation was still in flight. Not user-visible, but it is not a pure extraction.

## Tests

- New `xtream-refresh-flow.service.spec.ts` pins the reporter contract: re-entry after confirmation, busy-before-reset ordering, progress forwarding under one run, abort vs error, and a reporter that implements no paint hook.
- Both existing guard-reset regression tests were kept meaningful, not just green: removing the reset from the shared flow fails **exactly two** tests, one per entry point. Mutation-checked the new spec the same way (dropping `isBusy`, moving `begin()` after the reset).
- The sources spec's abort test counted a fixed number of microtasks to reach the delete; the shared flow adds one await to that path, so it now waits on a deterministic signal, matching the sibling spec's convention.

| Check | Result |
| --- | --- |
| `pnpm nx test playlist-shared-ui` | 98 passed (91 existing + 7 new), re-run after the rebase |
| `pnpm nx lint playlist-shared-ui` | clean, re-run after the rebase |
| `nx test workspace-shell-feature` / `workspace-dashboard-feature` | 51 + 144 passed |
| `nx build web` | passed |
| `electron-backend-e2e:e2e-ci--src/sources.e2e.ts` | 4 passed — covers this refresh in the real app |

Greptile reviewed the identical tree on #1430 and returned 5/5 with no findings.

## Docs

Updated both places that recorded the duplication as a fact: the "Explicit reset" call-site list in `docs/architecture/host-connectivity-guard.md` and the matching sentence in `CLAUDE.md`. Both now describe one flow with two entry points, and say a third should pass a reporter rather than copy the sequence.

No `.changes/` note — pure refactor, hence the `no-release-note` label.


4gray:
@codex review

4gray:
Fixed in 516530098 — thanks, this one is real.

The two entry points genuinely cannot see each other, and `XtreamPendingRestoreService.set()` overwrites unconditionally, so the second run does destroy the first run's parked favorites/history/hidden categories/playback positions.

One correction to the framing: the race predates this PR. On `master` the guards are already independent — `PlaylistRefreshActionService` checks its own `isRefreshing()` signal while `RecentPlaylistsComponent` checks its own per-row pending sets — so neither entry point could ever have fixed it alone. Extracting the flow is what makes it fixable once, which is why the fix lands here rather than in a follow-up.

`XtreamRefreshFlowService` now holds the set of playlists with a run in flight, refuses a second run *before* the connectivity-guard reset, and releases it in the same `finally` that ends the reporter so a finished run cannot strand a playlist. Covered by a new test that drives two independent reporters at one playlist; removing the guard fails it.

@greptileai

4gray:
@codex review

4gray:
Yes — and now stated rather than implied, in 3f7034a2f.

It was already covered: `sendIpcEvent` is only ever called by `resetHostConnectivityGuard`, so asserting it ran once proved the refused call never reached the reset. But you're right that a call count makes the reader infer the invariant, so the spec now asserts the recorded sequence directly — the refused call must add nothing at all, leaving exactly `[ipc:CONNECTIVITY_GUARD_RESET, delete]`.

I also ran your hypothetical rather than reasoning about it. Moving both the in-flight check and its insertion below the reset:

```ts
await resetHostConnectivityGuard(this.dataService, item.serverUrl);
if (this.activeRefreshPlaylistIds.has(item._id)) { return; }
this.activeRefreshPlaylistIds.add(item._id);
```

fails exactly this test and nothing else. (My first attempt at that mutation was wrong — I left `add()` above the check, so every run short-circuited on its own id and half the suite went red, which proves nothing. The corrected one is the single-test failure above.)

Worth naming why the ordering matters beyond "earlier is safer": the reset re-opens a host after the connection failures that led here, so a run that is not allowed to proceed must not clear that evidence on its way out.

@greptileai

PR #1433, opened 2026-08-13
refactor(portals): split the Stalker collection detail component

4gray:
## Why

`StalkerCollectionDetailComponent` was **529 non-blank lines**, well over the repo's 400-line production maximum (CLAUDE.md, "TypeScript File Size Rule"). It passed lint only because it sat in `tools/eslint/max-lines-baseline.mjs` — a list CLAUDE.md says must only ever shrink.

#1422 added ~22 lines to it for the "View in portal" handoff. That PR did not create the problem, but it made it worse, and the split was deliberately kept out to avoid widening its diff. This is that follow-up, on top of current master.

## What

Four cohesive units move into siblings in the same folder:

| File | Non-blank | Responsibility |
| --- | --- | --- |
| `stalker-collection-detail.component.ts` | **529 → 347** | Template, DI, effects, `prepareDetail` orchestration, view-in-portal |
| `stalker-collection-playback.controller.ts` | 166 | Playback ownership, inline playback, external-player fallback |
| `stalker-collection-detail-mode.ts` | 84 | Pure item / detail-mode / category resolution |
| `stalker-collection-favorites.controller.ts` | 67 | Favorites resource, sync, toggle |
| `stalker-collection-store-snapshot.ts` | 38 | Store snapshot capture/restore |

Both controllers follow the existing `StalkerVodPlaybackController` model — plain classes taking a config object, constructed in a component field initializer — rather than Angular DI services. That is deliberate: the playback controller needs the component's `item` input signal, and the favorites controller's `rxResource` needs an injection context. A field initializer supplies both for free, where a DI service would have needed an extra effect to feed the input across.

**Public API is unchanged.** `inlinePlayback`, `playbackSessionKey`, `selectedVodPlaybackPosition`, `isSelectedVodFavorite` and `portalFavorites` are re-exported signal *references*, so neither the template nor the spec sees the split.

## Validation

- `pnpm nx test portal-stalker-feature` → **272 passed / 24 suites**, exit 0. Identical count on unmodified `HEAD`, so nothing was lost or silently skipped.
- `pnpm nx lint portal-stalker-feature` → exit 0. The 12 remaining warnings are pre-existing, all in `stalker-search` / `stalker-series-view`; **zero** findings in any changed file.
- `pnpm run typecheck:web` → clean.
- Baseline regenerated with `node tools/eslint/generate-max-lines-baseline.mjs`: exactly one line removed, nothing added.
- Prettier-clean on all five files.

The spec is **untouched** and its 11 tests pass. (The suite carries a pre-existing worker-teardown flake — "worker process has failed to exit gracefully" — which reproduces on unmodified `HEAD`; failing runs report zero failing tests and zero failing suites. Not introduced here.)

## Docs

Moving `resolveDetailMode` / `resolveSelectedCategory` out of the class orphaned three doc references that named them as class methods. Updated in `collection-detail-portal-navigation.ts` (the two "Mirrors …" comments guarding the duplicated Stalker section rule) and `CLAUDE.md`. No other doc mentioned them.

## Release note

None — pure refactor, no behavior change. Carries `no-release-note` per CLAUDE.md, since the diff touches `libs/**`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1232,15 +1232,17 @@ engine` (restart required) or
   section-only route: an Xtream item without a resolvable category and positive
   item id keeps the action hidden rather than promising a jump to the title and
   landing in a list.
-- Stalker section resolution mirrors
-  `StalkerCollectionDetailComponent.resolveDetailMode()` and must not be
+- Stalker section resolution mirrors `resolveStalkerCollectionDetailMode()`
+  (`libs/portal/stalker/feature/src/lib/stalker-collection-detail-mode.ts`) and
+  must not be
   simplified to `item.contentType`: `extractStalkerItemType()` reports `series`
   for embedded `series[]` snapshots and lazy Ministra VOD `is_series` items, but
   both belong in the VOD catalog — the lazy season/episode fetch in
   `StalkerCatalogFacadeService.selectItem()` is gated on the VOD content type, so
   a `/series` route leaves the detail unable to load episodes. The virtual
   `series` category is normalized to `vod` the same way
-  `resolveSelectedCategory()` does. Stalker also carries `stalkerReturnTo`.
+  `resolveStalkerCollectionSelectedCategory()` does. Stalker also carries
+  `stalkerReturnTo`.
 - Unlike the download handoff this bridge does NOT pass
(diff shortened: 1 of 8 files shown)


4gray:
@codex review

PR #1434, opened 2026-08-13
test(playback): drive embedded MPV controller specs on a fake clock

4gray:
## Problem

`pnpm nx test ui-playback --skip-nx-cache` intermittently failed **2 tests of `embedded-mpv-session-controller.spec.ts` — a different pair each run**. E.g. one run failed *"loads support and unsubscribes from session updates on destroy"* + *"ignores a late startup rejection after teardown…"*, the next failed *"starts a session, forwards matching updates…"* + *"sets an error session when Electron cannot create playback"*.

## Root cause

The spec's `waitFor` helper polled **real timers against a `Date.now() + 1_000` wall-clock deadline**. Under load from parallel Jest workers, the controller's async startup chain (support load → `waitForStartupPaint` → session create → update subscription) did not settle inside that budget, and whichever tests lost the race failed. The controller itself is not at fault — only the spec's flushing strategy.

Reproduced deterministically by running the suite at `--maxWorkers=32` on a 10-core machine; every failure was `Timed out waiting for …` thrown by `waitFor` itself.

## Fix

`jest.useFakeTimers()` in `beforeEach` / `useRealTimers()` in `afterEach`, and `waitFor` now bounds itself by **drain rounds instead of elapsed time**, so machine load can no longer change the outcome. Added a `flush()` helper for the negative assertions that have no condition to poll.

**Assertions and behavior under test are unchanged** — only the flushing mechanism.

### One non-obvious detail

Each round advances **1ms, never 0ms**. `waitForStartupPaint()` nests rAF inside rAF, and a zero-delay timer scheduled from *inside* a timer callback is clamped to the next millisecond. A `advanceTimersByTimeAsync(0)` fires the outer hop and **strands the inner one forever** — that would have converted the flake into a hang to Jest's 5s timeout. Verified empirically with a throwaway probe before committing to the approach:

| advance | nested rAF chain |
|---|---|
| `0` | `["raf1"]` — never completes |
| `1` | `["raf1","raf2","resolved"]` |

## Scope

The task named one file, but the two siblings — `.dpr.spec.ts` and `.lifecycle.spec.ts` — carry a **byte-identical `waitFor`** and were demonstrated flaking under the same repro. They are included because `--testPathPatterns` is ignored by `tools/testing/run-web-esm-lib-tests.mjs` (it passes `--runTestsByPath` with every spec, which takes precedence), so any verification run of this project executes all three; leaving them would have made the fix unverifiable. The pattern now appears nowhere else in the repo.

## Verification

| condition | before | after |
|---|---|---|
| `--maxWorkers=32` (repro) | failed 2/2 runs | **5/5 runs, 975/975** |
| `pnpm nx test ui-playback --skip-nx-cache` ×5 | passed (flake needs load) | **5/5 runs, 975/975** |
| three specs solo | ~62s est. | 18/18 in 4.2s |

ESLint clean on all three files. The `"A worker process has failed to exit gracefully"` warning also disappeared — fake timers stop the stalled tracker's 30s timeout leaking into the real timer queue.

## Notes

- **No release note**: test-only change; `.spec.ts` is auto-exempt in `tools/release/check-release-note-gate.mjs:31`, so no `no-release-note` label is required either.
- **No doc update**: CLAUDE.md lists isolated test-only changes as a skip case.
- Two *unrelated* specs (`VodDetailsComponent`, `WebPlayerViewComponent recovery integration`) also failed under `--maxWorkers=32` before this change, but via Jest's own 5000ms timeout — a different mechanism, not investigated here. They were green across all 10 post-fix runs.

### DIFF
--- a/libs/ui/playback/src/lib/embedded-mpv-player/embedded-mpv-session-controller.dpr.spec.ts
+++ b/libs/ui/playback/src/lib/embedded-mpv-player/embedded-mpv-session-controller.dpr.spec.ts
@@ -50,6 +50,9 @@ describe('EmbeddedMpvSessionController devicePixelRatio watch', () => {
     let mediaQueries: FakeMediaQueryList[];
 
     beforeEach(() => {
+        // See `waitFor`: the startup chain is drained on a virtual clock so a
+        // loaded machine cannot change the outcome.
+        jest.useFakeTimers();
         electron = {
             platform: 'win32',
             getEmbeddedMpvSupport: jest
@@ -110,6 +113,7 @@ describe('EmbeddedMpvSessionController devicePixelRatio watch', () => {
         TestBed.resetTestingModule();
         delete (window as unknown as { electron?: unknown }).electron;
         delete (window as unknown as { matchMedia?: unknown }).matchMedia;
+        jest.useRealTimers();
         jest.restoreAllMocks();
     });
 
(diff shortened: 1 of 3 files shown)


4gray:
@codex review

PR #1435, opened 2026-08-13
fix(portals): return from a portal handoff without losing the collection view

4gray:
Follow-up to #1422, addressing the P2 Codex raised on it after it merged.

## Problem

`View in portal` put `stalkerReturnTo` (the collection URL) in the navigation state, and the portal detail's back affordance re-navigated there with `navigateByUrl()`. That starts a **stateless** history entry — but the collection page keeps its active tab, scope and open inline detail only in `window.history.state` (`collectionViewState` / `openCollectionDetailItem`).

Traced end to end rather than taken on faith:

1. `stalkerReturnTo: router.url` — set by the collection handoff
2. `onVodBack()` / `goBack()` → `router.navigateByUrl(returnTo)` — fresh entry, no state
3. `getCollectionViewState(window.history.state)` → `null`, and `selectedContentType` falls back to **`'live'`** (`unified-collection-page.component.ts:146`)

So: open a Stalker movie from Global favorites on the **Movies** tab → View in portal → back → you land on the **Live TV** tab with the title closed. The re-navigation also pushes an entry, so browser Back then goes forward into the portal again instead of further back.

## Fix

The handoff additionally sets `stalkerReturnByHistory`, and both Stalker back handlers step back one history entry instead of re-navigating.

That entry is exactly the one the handoff itself pushed — `stalkerReturnTo` is read off `window.history.state`, so it is only ever present on that entry — and `pushInlineDetailState()` spreads the previous state forward, so the entry underneath carries both `collectionViewState` and `openCollectionDetailItem`. One step back therefore restores the collection precisely as the user left it, with the title still open, and adds nothing to the history stack.

The flag is set **only** by `getUnifiedCollectionDetailNavigation()` and only alongside `returnTo`, so the dashboard handoff (`returnTo: '/workspace/dashboard'`) and every other `stalkerReturnTo` caller keep their existing re-navigating behaviour. The reader accepts an exact boolean `true`, nothing looser.

## Scope note

`stalker-collection-detail.component.ts` is deliberately **untouched** — the flag is added in the shared navigation builder — because a separate task is currently splitting that file for its `max-lines` overage. This PR should rebase cleanly under it either way.

`stalker-series-view.component.spec.ts` is already at ~1230 effective lines against the 1200 test limit, so the new coverage went into a focused `stalker-series-view.back-navigation.spec.ts`, matching the sibling specs in that folder.

## Tests

7 new tests:

- builder: sets the flag with `returnTo`, omits it without, and rejects non-`true` values / absent state
- `StalkerCatalogDetailComponent.onVodBack()` and `StalkerSeriesViewComponent.goBack()`: each covers history-step, plain re-navigate, and no-target

The two history-step tests were verified to **fail on the previous behaviour** (reverted each handler in turn, got exactly the one expected failure, restored) so they cannot pass vacuously. The existing `stalker-collection-detail` assertion was updated for the added state key.

`pnpm nx run-many --target=test --projects=portal-shared-util,portal-stalker-feature` → **480 passed** (202 + 278). Lint clean, `pnpm run release:notes:validate` passes.

### DIFF
--- a/.changes/portals-view-in-portal-back.md
+++ b/.changes/portals-view-in-portal-back.md
@@ -1,0 +1,8 @@
+---
+type: fix
+area: portals
+---
+
+Going back from a Stalker title you reached with "View in portal" now returns
+you to exactly where you were — the same favorites or history tab, with the
+title still open — instead of dropping you on the Live TV tab.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1257,7 +1257,16 @@ engine` (restart required) or
   a `/series` route leaves the detail unable to load episodes. The virtual
   `series` category is normalized to `vod` the same way
   `resolveStalkerCollectionSelectedCategory()` does. Stalker also carries
-  `stalkerReturnTo`.
+  `stalkerReturnTo` plus
+  `stalkerReturnByHistory`, and the portal detail's back affordance
+  (`StalkerCatalogDetailComponent.onVodBack()`,
+  `StalkerSeriesViewComponent.goBack()`) honours the latter by stepping back
+  one history entry instead of calling `navigateByUrl()`. The collection's
+  active tab, scope and open inline detail live only in `window.history.state`
+  (`collectionViewState` / `openCollectionDetailItem`), so re-navigating would
+  reopen it on the default `live` tab and leave the portal page one browser
+  Back away. Only this builder sets the flag, so the dashboard handoff and any
+  other `stalkerReturnTo` caller keeps re-navigating.
 - Unlike the download handoff this bridge does NOT pass
   `detailPresentation: 'provider-only'` — the item exists in the provider
   catalog, so the full normal detail (downloads included) is wanted.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1271,10 +1271,12 @@ engine` (restart required) or
   after Back + browser Forward the same entry can host a different title, whose
   back affordance must just close it. A stale marker suppresses the whole
   return contract, and honouring it retires both keys from the entry so a
-  browser Forward cannot replay them for a reopened title. The comparison
-  identity follows `extractStalkerItemId()`'s field order
-  (`id ?? stream_id ?? series_id ?? movie_id`) — a collection row may carry
-  only `movie_id`/`series_id`. Only this builder sets the marker, so the
+  browser Forward cannot replay them for a reopened title. The identity is
+  restricted to what `buildStalkerSelectedVodItem()` preserves (`id ??
+stream_id`); it drops `series_id`/`movie_id`, so a row identified only by
+  those gets no marker at all and the handoff falls back to re-navigating
+  rather than comparing against an identity the opened detail cannot report.
+  Only this builder sets the marker, so the
   dashboard handoff and any other `stalkerReturnTo` caller keeps
   re-navigating.
 - Unlike the download handoff this bridge does NOT pass
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1265,7 +1265,12 @@ engine` (restart required) or
   active tab, scope and open inline detail live only in `window.history.state`
   (`collectionViewState` / `openCollectionDetailItem`), so re-navigating would
   reopen it on the default `live` tab and leave the portal page one browser
-  Back away. Only this builder sets the flag, so the dashboard handoff and any
+  Back away. The marker carries the handed-off item's identity, not a bare
+  `true`: `openStalkerItem` is consumed on arrival while the return keys stay
+  on the entry, and a Stalker detail opens in place without pushing one — so
+  after Back + browser Forward the same entry can host a different title, whose
+  back affordance must just close it. A stale marker suppresses the whole
+  return contract. Only this builder sets it, so the dashboard handoff and any
   other `stalkerReturnTo` caller keeps re-navigating.
 - Unlike the download handoff this bridge does NOT pass
   `detailPresentation: 'provider-only'` — the item exists in the provider
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1274,11 +1274,13 @@ engine` (restart required) or
   browser Forward cannot replay them for a reopened title. Leaving with the
   browser's own Back runs no affordance, so `CategoryContentViewComponent`
   also retires the contract whenever it lands on the entry with no handoff
-  item and no open detail. The identity is
+  item and no open detail. That retirement is gated on the marker, so a plain
+  `stalkerReturnTo` caller such as the dashboard handoff is unaffected. The identity is
   restricted to what `buildStalkerSelectedVodItem()` preserves (`id ??
-stream_id`); it drops `series_id`/`movie_id`, so a row identified only by
-  those gets no marker at all and the handoff falls back to re-navigating
-  rather than comparing against an identity the opened detail cannot report.
+stream_id`); it drops `series_id`/`movie_id`, so the builder pins the
+  resolved id onto the handoff state item when the raw row carries neither —
+  those rows then get the same history return instead of degrading to a
+  re-navigation that resets the collection's tab.
   Only this builder sets the marker, so the
   dashboard handoff and any other `stalkerReturnTo` caller keeps
   re-navigating.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1270,8 +1270,13 @@ engine` (restart required) or
   on the entry, and a Stalker detail opens in place without pushing one — so
   after Back + browser Forward the same entry can host a different title, whose
   back affordance must just close it. A stale marker suppresses the whole
-  return contract. Only this builder sets it, so the dashboard handoff and any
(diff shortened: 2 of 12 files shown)


4gray:
@codex review

4gray:
Fixed in e794d0b80 — the finding is valid on the mechanism, and I verified each step rather than taking it on faith.

**Confirmed:** `clearNavigationStateKeys()` in `openStalkerItemFromNavigationState()` removes `openStalkerItem` but deliberately leaves the return keys; `StalkerCatalogFacadeService.selectItem()` returns `null`, so `onItemClick()` opens a Stalker detail **in place** without pushing a history entry. After Back + browser Forward, the same entry can therefore host a different title while still carrying the marker.

**Fix:** `stalkerReturnByHistory` now carries the handed-off item identity instead of a bare `true`, and a marker that does not match the currently open title is treated as stale — it suppresses the *whole* return contract, so back just closes the detail. Suppressing both paths matters: gating only `location.back()` would have left the equally stale `stalkerReturnTo` to fire `navigateByUrl()` and produce the same unexpected exit. Lazy episode ids (`parent:child`) normalize to the parent, mirroring the catalog view's own `stalkerItemIdentity()`.

Three new tests cover it (bound match, parent-id match, stale mismatch on both back handlers). I verified they fail without the binding — reverted `isStalkerReturnByHistoryFor` to an unbound check and got exactly those failures, then restored.

**Scope note:** the underlying staleness is pre-existing for `stalkerReturnTo` on the dashboard handoff (`returnTo: '/workspace/dashboard'`), which sets no marker and is untouched here. This PR makes the contract it introduces exact rather than rewriting that one.

`portal-shared-util` 202 + `portal-stalker-feature` 280 = 482 tests green; lint clean.

@greptileai

4gray:
@codex review

4gray:
Both observations acted on in c2b0e790c — they were worth fixing rather than noting.

**Identity divergence.** Rather than confirming the two stay aligned by inspection, they are now one function: `CategoryContentViewComponent.stalkerItemIdentity()` delegates to the shared `normalizeStalkerHandoffIdentity()`, so the marker binding and the catalog view cannot drift.

**Handler duplication.** Extracted `resolveStalkerBackNavigation(state, selectedItemId)`, which returns `history-back` / `navigate` / `none`; both handlers just apply the result. The precedence rule now has direct unit coverage (4 cases) instead of only being exercised through two components, and a third consumer would inherit it for free.

You were right that the ordering was already correct — the clears run unconditionally before the gate, and that is preserved.

`portal-shared-util` 206 + `portal-stalker-feature` 280 + `portal-catalog-feature` 14 = 500 tests green; lint and Prettier clean.

@greptileai

4gray:
@codex review

4gray:
Both P2s from the `e794d0b80c` review are valid and fixed in 423a75629. (Note the review predated c2b0e790c, but the refactor only moved that logic, so both still applied.)

**1. ID-shape mismatch — this one broke the feature outright.** Confirmed: the marker is built from `item.stalkerId`, i.e. `extractStalkerItemId()` = `id ?? stream_id ?? series_id ?? movie_id`, while the comparison read `item.id` alone. For a row carrying only `movie_id`/`series_id` the identity was empty, so `returnsHere` was false — and because a marker existed the handler returned without stepping back *or* using `stalkerReturnTo`. Back stopped doing anything at all. That was a regression introduced by the binding itself, not a pre-existing wart. `stalkerHandoffIdentityOf()` now mirrors that exact field order.

**2. Same-title replay.** Also correct: binding fixed a *different* title on the entry but still matched the original one reopened from the catalog. Honouring the marker now calls `consumeStalkerReturnMarker()`, retiring both `stalkerReturnByHistory` and `stalkerReturnTo` from the entry before `location.back()` — the handoff is genuinely one-shot, so a Forward-replay finds nothing to fire.

5 new tests, each verified to fail without the fix (reverted both changes, got exactly those 5 failures, restored): resolver-level `movie_id`/`series_id`/`stream_id` matching, plus per-handler coverage for the retired keys and for a `movie_id`-only selection.

`portal-shared-util` 207 + `portal-stalker-feature` 284 + `portal-catalog-feature` 14 = 505 tests green; lint and Prettier clean.

@greptileai — for the record on your last note: yes, `{ kind: 'none' }` is deliberate. Both handlers clear the player/selection unconditionally before branching, so `none` means "the detail closes, nothing navigates" — never a stranded open detail.

4gray:
@codex review

4gray:
Valid, and it invalidates my previous fix — corrected in 4df6ba003.

Confirmed: `buildStalkerSelectedVodItem()` builds `id: toStringOrFallback(item.id ?? item.stream_id)` and carries no `series_id`/`movie_id`. Every opened detail goes through it, so widening the lookup to `extractStalkerItemId()`'s four fields ran *after* the data was already gone — the comparison still could not match, and the back affordance still did nothing for those rows.

Worth stating plainly: **my regression test for that fix was wrong**, and that is why the round shipped broken. It set `selectedItem` to `{ movie_id: '42' }` directly — a shape normalization can never produce — so it passed against a pipeline that was still failing. Both handler specs now build the selection through `buildStalkerSelectedVodItem()` so they exercise the real shape.

**The fix:** bind the marker to the identity the detail can actually report (`id ?? stream_id`). A row identified only by `movie_id`/`series_id` now gets **no marker at all**, so the handoff falls back to plain `stalkerReturnTo` re-navigation — the pre-existing behaviour, which works (it only loses the collection tab) rather than stranding back entirely. Fail-safe instead of fail-silent.

Verified the new builder test catches the regression: re-bound the marker to the wide id set and it failed, restored and it passes.

`portal-shared-util` 208 + `portal-stalker-feature` 284 + `portal-catalog-feature` 14 = 506 green; lint and Prettier clean.

@greptileai

4gray:
@codex review

4gray:
Valid — fixed in ace10995f. This was the residual hole: the marker was only ever retired inside the two in-app `history-back` branches, so leaving via the browser's own Back left it live, and reopening the *same* title after a Forward matched the identity.

**Fix:** `CategoryContentViewComponent.openStalkerItemFromNavigationState()` — which already owns arrival-state consumption — now retires the contract when it lands on the entry with **no handoff item and no detail open**. That is precisely the post-handoff state: `openStalkerItem` was consumed on arrival, so a marker without it means the handoff is over and anything opened from the list afterwards is a fresh selection.

The `!this.selectedItem()` guard matters: during arrival itself the detail is already selected, so a re-emission of `paramMap` cannot retire a contract that is still needed. Both directions are covered by tests — one asserts the stale contract is stripped, the other that arrival keeps it.

Verified the new test catches it: removed the retirement and it failed, restored and it passes.

Also folded in @greptileai's non-blocking note — both handlers now state explicitly that closing the detail is unconditional and a `none` decision only suppresses the navigation.

`portal-shared-util` 208 + `portal-stalker-feature` 284 + `portal-catalog-feature` 16 = 508 green; lint and Prettier clean.

@greptileai

4gray:
@codex review

4gray:
Right again, and the better design — fixed in 8fa09003d.

My "graceful degradation" was not graceful: falling back to `navigateByUrl()` for alternate-id rows reproduced **exactly** the defect this PR exists to fix (stateless entry → collection reopens on its default tab, title closed). Choosing the old bug as a fallback is not a safe default when the whole PR is about removing that bug.

**Fix, per your second suggestion:** the builder pins the resolved id onto the handoff state item when the raw row carries neither `id` nor `stream_id`. `buildStalkerSelectedVodItem()` then reports an identity the marker can bind to, so those rows get the same history return as every other one. An existing `id` is never overwritten (covered by its own test).

**Separately, I found a scope violation of my own** while re-reading the previous round: the arrival-side retirement cleared `stalkerReturnTo` unconditionally, so it also affected the dashboard handoff — which I had claimed throughout this PR was untouched. It is now gated on the marker's presence, with a test asserting a plain `stalkerReturnTo` state survives.

Both new behaviours verified against their inverse (removed the pinning → the alternate-id test failed; unscoped the retirement → the dashboard test failed; restored both).

`portal-shared-util` 209 + `portal-stalker-feature` 284 + `portal-catalog-feature` 17 = 510 green; lint and Prettier clean.

@greptileai

4gray:
@codex review

4gray:
Both bots came back clean on `8fa09003d` — Codex "didn't find any major issues", Greptile "everything checks out".

Pushed one last commit (0f005ab94) taking @greptileai's optional nit: the optional chain and `?? ''` fallback in `normalizeStalkerHandoffIdentity` really were unreachable rather than type-required. I checked `tsconfig.base.json` first — with `noUncheckedIndexedAccess` enabled that fallback would have been mandatory and removing it would have broken the build. The workspace does not set it, so `split(':')[0]` is typed `string` and the guard is dead. Behaviour unchanged, 510 tests still green.

Worth noting for anyone reading the CI tab: this branch's runs are being **cancelled** by each successive push (concurrency group), not merely queued — which is why the completed runs lag several commits behind. The final SHA is the one that matters, so I am letting this one run out.

@greptileai

4gray:
@codex review

Issue #1338, opened 2026-08-02
Search input bug: space deletion and flickering search results


okasha007:
**Describe the bug**
When typing in the search bar, spaces are automatically deleted/ignored while typing, making multi-word searches very difficult. Additionally, search results flicker, appear briefly, and then disappear, leaving a blank or frozen page.

**To Reproduce**
1. Go to the search bar in IPTVnator.
2. Try to type a query containing spaces (e.g., "Bein Sports").
3. Notice that the space is automatically removed while typing and the search results UI glitches/disappears.

**Expected behavior**
Smooth typing in the search input without auto-deleting spaces, and stable search results (or adding a debounce / option to disable live search-as-you-type).

**Screenshots**
N/A

**Desktop (please complete the following information):**
- PWA or Electron/Tauri application: Electron
- OS: Windows
- Browser: N/A
- Version: [Latest]
- Feature: M3U playlist / Xtream Code

4gray:
Fixed in v0.23.0: search preserves spaces and no longer resets while typing ([#1474](https://github.com/4gray/iptvnator/pull/1474), [#1432](https://github.com/4gray/iptvnator/pull/1432)). Please reopen if this still occurs on v0.23.0 or later.

PR #1474, opened 2026-08-23
fix(search): keep a trailing space when the trimmed q echo lands after the debounce

4gray:
Fixes the residual part of #1338: on a query-search page, typing a space and pausing past the 350 ms debounce deleted the space — continuing to type "Bein Sports" collapsed into "BeinSports".

## Root cause

The guard from #1432 only held while a debounce was still pending (`searchDebounceTimeoutId !== null`). Once the debounce fired, the sync effect wrote the trimmed term to the URL (`q=Bein`, replaceUrl), and on `NavigationEnd` the echo ran `setSearchState("Bein")` — the header input is one-way bound to `searchQuery`, so the trailing space the user just typed was wiped.

## Fix

Two coordinated changes in `WorkspaceShellSearchSyncService`:

- **`applySearchQuery` now always applies the trimmed term.** The URL sync and portal stores only ever act on the trimmed form anyway; an untrimmed applied term is exactly what made the router echo of our own `q` look like foreign search intent. The box (`searchQuery`) keeps what was actually typed, whitespace included.
- **The echo guard no longer requires a pending debounce.** A same-page, imperative navigation whose `q` equals the applied term is our own echo (or an unrelated query-param write) whether or not a keystroke is still waiting. Back/forward stays authoritative: the `trigger === 'imperative'` check is untouched, so popstate always re-applies what the history entry carries.

## Tests

- New regression spec (trailing-space pause) verified to fail on the old code: 1 failed / 8 passed with the service reverted.
- New popstate spec proves history re-apply still wins after typing settles.
- Facade spec router mock gained `lastSuccessfulNavigation` — the widened guard now evaluates it on the constructor's initial sync.
- `pnpm nx test workspace-shell-feature` (155 tests) and `pnpm nx lint workspace-shell-feature` pass.

Release note: `.changes/search-trailing-space-typing.md` (`type: fix`, `area: search`, issue 1338).

### DIFF
--- a/.changes/search-trailing-space-typing.md
+++ b/.changes/search-trailing-space-typing.md
@@ -1,0 +1,9 @@
+---
+type: fix
+area: search
+issues: [1338]
+---
+
+Typing a space in the portal search box and pausing briefly no longer deletes
+the space — continuing to type "Bein Sports" no longer collapses into
+"BeinSports".
--- a/docs/architecture/workspace-shell.md
+++ b/docs/architecture/workspace-shell.md
@@ -165,12 +165,18 @@ Search is shell-owned and route-aware:
 8. The URL is authoritative for the search box only when it carries search
    intent. `WorkspaceShellSearchSyncService` re-reads `q` on every
    `NavigationEnd`, but an **app-initiated** navigation that stays on the same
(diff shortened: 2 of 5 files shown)


4gray:
@codex review

PR #1517, opened 2026-09-03
fix(stalker): keep live playback when switching the ITV/radio category

4gray:
## Summary
- In Stalker portals, picking another Live TV or radio category in the sidebar stopped the channel that was playing. Xtream live (#936) and M3U groups already keep playback across a category/group switch; Stalker now does too.
- Root cause: `onStalkerCategoryClicked` in the workspace context panel called `clearSelectedItem()` for every section, and `StalkerLiveStreamLayoutComponent` renders its player only while the store has a `selectedItem`.

## Changes
- `workspace-context-panel.component.ts`: for `itv`/`radio` the handler only sets the category (and resets the page) and returns before `clearSelectedItem()`. VOD/series clicks still close the open detail before navigating to the list route.
- `stalker-live-stream-layout.component.ts`: the category-change reset effect keeps clearing list state (legacy-paged channels, page, row EPG previews) but no longer wipes the playing channel's short-EPG fallback or invalidates a fallback load in flight. That reset moved into a separate effect keyed on the content type, so leaving the section (`itv` ↔ `radio`, where the route session clears the selection) still drops it.
- Docs: `stalker-portal.md` (Live TV and Radio), `workspace-shell.md` (Context Panel Rules), `stalker-epg.md` (Fallback Behavior).
- Release note: `.changes/stalker-keep-live-playback-on-category-switch.md` (`type: fix`, `area: stalker`).

## Testing
- [x] `workspace-context-panel.component.spec.ts`: ITV category click preserves the selection, radio test now asserts no clear, VOD click still clears and navigates (2 of these failed on the old code).
- [x] `stalker-live-stream-layout.component.spec.ts`: player and the playing channel's EPG fallback survive a category change, an in-flight fallback load finishes across the switch, leaving ITV for radio still drops the fallback and its loading flag (2 failed on the old code).
- [x] `apps/web-e2e/src/stalker.e2e.ts`: new `@stalker ITV playback survives a category switch` — green in chromium, webkit and firefox; the whole Stalker suite passed via `pnpm nx run web-e2e:e2e-ci--src/stalker.e2e.ts`.
- [x] `pnpm nx test portal-stalker-feature` (live layout specs), `pnpm nx test workspace-shell-feature` (context panel), `pnpm nx run-many -t lint -p portal-stalker-feature workspace-shell-feature web-e2e`, `pnpm run release:notes:validate`.
[...]

PR #1522, opened 2026-09-03
test(stalker): cover radio playback surviving a category switch in web E2E

### LINKED ISSUE #1521: E2E: cover Stalker radio playback surviving a category switch
## Context

#1517 made Stalker Live TV and radio keep playing when the user picks another category in the shell context panel. The web E2E added there (`@stalker ITV playback survives a category switch` in `apps/web-e2e/src/stalker.e2e.ts`) covers ITV only. Radio goes through the same `onStalkerCategoryClicked` early return and is covered by a unit test in `workspace-context-panel.component.spec.ts`, but there is no end-to-end proof that the inline audio player stays mounted across a radio category switch.

## Task

Add a `@stalker` web E2E scenario next to the existing radio test (`@stalker radio — stations use the inline audio player without EPG`):

1. Add the portal, open the Radio tab, pick the first real category, wait for `[data-test-id="channel-item"]`.
2. Click a station and wait for `app-audio-player` to be visible.
3. Click another category, wait for the sidebar `.category-title` to change and for the station list to re-render.
4. Assert `app-audio-player` is still visible.

## Notes

- The mock portal (`apps/stalker-mock-server`) exposes radio categories in the default scenario (`00:1a:79:00:00:01`); check the existing radio test for the exact navigation and selectors.
- The ITV variant of this scenario is the template — mirror its "title changed" guard so the assertion cannot pass before change detection ran.
- Test-only change: no `.changes/` note needed (E2E paths are auto-exempt from the release-note gate).

4gray:
## Summary
- Follow-up to #1517: the web E2E there covers ITV only. Radio shares `StalkerLiveStreamLayoutComponent` and the same `onStalkerCategoryClicked` early return, and its inline `app-audio-player` is gated on the store selection just like the ITV player, so a regression would silence a station the user never switched away from.
- Adds `@stalker radio playback survives a category switch` to `apps/web-e2e/src/stalker.e2e.ts`, mirroring the ITV scenario: play a station, pick another category, wait for the sidebar `.category-title` to change (so the assertion cannot pass before change detection ran), then assert the audio player is still visible.

## Changes
- `apps/web-e2e/src/stalker.e2e.ts`: one new test next to the existing radio test. No production code touched.

## Testing
- [x] `pnpm nx run web-e2e:e2e-ci--src/stalker.e2e.ts` — whole Stalker suite green; the new scenario passed in chromium, webkit and firefox.
- [x] `pnpm nx lint web-e2e` — clean (pre-existing warnings only).
- Test-only change: E2E paths are auto-exempt from the release-note gate, so no `.changes/` note.

Closes #1521

### DIFF
--- a/apps/web-e2e/src/stalker.e2e.ts
+++ b/apps/web-e2e/src/stalker.e2e.ts
@@ -552,6 +552,44 @@ test('@stalker radio — stations use the inline audio player without EPG', asyn
     expect(epgRequests).toHaveLength(0);
 });
 
+test('@stalker radio playback survives a category switch', async ({ page }) => {
+    // Radio shares the ITV layout and the same context-panel handler that
+    // used to clear the selected item on every category click (#1517). The
+    // inline audio player is gated on that selection, so a category switch
+    // silenced the station the user never switched away from.
+    await addStalkerPortal(page);
+
+    await page.getByRole('link', { name: /radio/i }).click();
+    await page.waitForURL(/stalker.*radio/);
+
+    const categories = page.locator('.category-item');
+    await expect(categories.nth(2)).toBeVisible({ timeout: 10_000 });
+    await categories.nth(1).click();
+
+    const sidebar = page.locator('app-stalker-live-stream-layout .sidebar');
+    const sidebarTitle = sidebar.locator('.category-title');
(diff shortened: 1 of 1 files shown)


4gray:
@codex review

PR #1532, opened 2026-09-05
fix(dashboard): read cached EPG timestamps as unix seconds

4gray:
## Summary

- `epgTimestampMs` in `libs/workspace/dashboard/feature/src/lib/rails/dashboard-live-epg.utils.ts` returned `EpgProgram.startTimestamp` / `stopTimestamp` verbatim, but those fields are unix **seconds** everywhere else in the app (`getProgramTimeMs` in `libs/ui/epg/src/lib/epg-program.utils.ts` multiplies by 1000). A row that carried them would have put the dashboard's "now playing" range in 1970 and pinned the progress bar at 100%.
- The bug is latent today: the backend `GET_CURRENT_PROGRAMS_BATCH` lookup (`transformDbRowToEpgProgram` in `epg-query.service.ts`) never sets those fields, so the ISO fallback always ran. This closes the gap before any data path supplies them.
- Fix mirrors `getProgramTimeMs`: a positive finite timestamp is scaled by 1000; zero / negative / non-finite values are treated as absent and fall back to the ISO string.

## Tests

- New `dashboard-live-epg.utils.spec.ts`: feeds seconds-based `startTimestamp`/`stopTimestamp` next to deliberately different ISO strings and asserts the formatted `20:00 – 21:00` range and 50% progress, plus the zero/NaN fallback, plain ISO input, clamping, unresolvable ranges and the null paths of `buildDashboardLiveEpgDetails`. Fixtures use local wall-clock dates so the assertions hold in any timezone.
- `workspace-dashboard-rails.component.spec.ts`: the existing fixture stored `Date.UTC(...)` milliseconds in the seconds fields and so encoded the old bug (4 cases failed against the fix). It now stores seconds and derives the millisecond clock values from them.

## Validation

| Check | Result |
| --- | --- |
| `pnpm nx test workspace-dashboard-feature` | 4 suites, 61 tests passing |
| `pnpm nx lint workspace-dashboard-feature` | passed |
| `pnpm run release:notes:validate` | valid |

## Release note

`.changes/dashboard-live-epg-timestamp-units.md`, `type: internal` — nothing changes on screen today. No docs update needed.


4gray:
@greptileai

4gray:
@codex review

PR #1533, opened 2026-09-05
fix(playback): hide VLC RC console on Windows

### LINKED ISSUE #871: VLC opens via console window after update
Description

Since updating the application to version 0.19.0, videos opened via the integrated video-selection player start VLC media player through a console window first.

Expected Behavior

VLC should start directly without opening a console window.

Actual Behavior

A console window appears first, and VLC is launched through it.

Steps to Reproduce
Open the application
Select VLC as video player
Select a video in a playlist
Start playback

Environment
OS: Windows 11 (64-bit)
Also occurred on another Setup with Windows 10 (64-bit)
App version: 0.19.0
VLC version: 3.0.23

Workaround

Start VLC manually and limit it to a single instance. New videos will then be added to running vlc playlist instead of launching a new instance+console.

Notes

This behavior did not occur in previous versions (e.g. 0.18.0, 0.17.0) with the same setup.

4gray:
VLC's managed RC interface opens a separate DOS console on Windows when progress tracking or instance reuse is enabled. Add `--rc-quiet` only to Windows launches with an allocated RC port, keeping TCP control, custom-argument ordering, and both no-RC retry paths intact. Retry filtering now matches the quiet flag exactly so URLs, headers, and titles containing that text survive. macOS/Linux and launches without RC keep their existing arguments.

Closes #871

Validation:
- `pnpm nx test electron-backend --maxWorkers=2`: all 158 suites / 1949 tests pass on the final code.
- VLC regression suites: 46 tests pass; five assertions failed on the old code because `--rc-quiet` was missing. Two additional regressions reproduced the retry filter dropping URLs/headers/titles containing `--rc-quiet`; both now pass with exact flag matching. Coverage includes progress-only/reuse-only RC, Windows/macOS/Linux, custom quiet flags, port-allocation failure, and spawn-error/exit-code-1 retries.
- `pnpm nx lint electron-backend`: passes with existing warnings; changed files also lint cleanly.
- `pnpm nx run electron-backend-e2e:e2e-ci--src/dash-clearkey.e2e.ts`: passes (1 test), including the Electron E2E build. This verifies the external-fallback UI with captured launch IPC, not a real VLC process.
- `pnpm run release:notes:validate` and `git diff --check`: pass.

Windows is unavailable in this macOS environment, so an actual Windows VLC console smoke test remains unperformed. The Windows argument tests mock the platform and child process; the issue's Windows 11 report and VideoLAN's Windows-only RC option corroborate the fix.

Added `.changes/playback-vlc-windows-console.md` and updated the external-player argument contract in `docs/architecture/embedded-inline-playback.md`.

### DIFF
--- a/.changes/playback-vlc-windows-console.md
+++ b/.changes/playback-vlc-windows-console.md
@@ -1,0 +1,7 @@
+---
+type: fix
+area: playback
+issues: [871]
+---
+
+VLC now opens without an extra console window on Windows when playback progress tracking or Reuse VLC instance is enabled.
--- a/apps/electron-backend/src/app/events/vlc-session.service.lifecycle.spec.ts
+++ b/apps/electron-backend/src/app/events/vlc-session.service.lifecycle.spec.ts
@@ -47,6 +47,7 @@ import { externalPlayerSessions } from './external-player-runtime';
 import { externalPlayerProcessTeardownGate } from './external-player-process';
 import { openVlcPlayer, shutdownVlcSession } from './vlc-session.service';
 
+const originalPlatform = process.platform;
 const spawnMock = spawn as unknown as jest.Mock;
 const streamUrl = 'https://example.com/stream.m3u8';
 const rcWrites: string[] = [];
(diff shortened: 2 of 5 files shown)


4gray:
Addressed the Codex retry-filter finding: both retry paths now remove only the exact managed `--rc-quiet` argument. URL, HTTP-option, and title values containing that substring are preserved, as are user-supplied flags. Two regression cases (spawn error and exit code 1) failed with the substring filter and now pass; all 46 VLC tests and changed-file lint pass.

@greptileai

@codex review

PR #1544, opened 2026-09-05
fix(stalker): let media transports control seek byte ranges

4gray:
## What changed

Remove the fixed `Range: bytes=0-` header from Stalker's direct-stream profile in both the renderer and Electron's fallback playback context. The media transport now chooses the byte range for each seek. Add regression coverage for foreign hosts, TLS downgrades, portal-owned streams, and resuming a CDN episode; update the playback header contract.

## Why

After continuing a Stalker episode in Embedded MPV, seeking backward could jump forward or reach EOF and advance to the next episode. Reproduced on v0.23 and current master: the static header overrides mpv's requested byte offset. A generated 20-minute MKV served over local HTTP reproduced the failure with the header; without it, resume at 13:00 and repeated seeks to 10:00, 10:00, and 5:00 landed exactly. The same correction was verified on the reported episode through CDP, including both ±10-second UI buttons.

## Release note

- [x] Added `.changes/stalker-seek-byte-ranges.md` and validated release notes.
- Updated `docs/architecture/stalker-portal.md` (Playback Header Contract).

## Checks

- [x] New regression assertions fail before the corresponding fix and pass afterwards.
- [x] `pnpm nx test portal-stalker-data-access --runInBand` — 524 tests passed.
- [x] `pnpm nx test electron-backend --runInBand --testPathPatterns='(stalker-playback-context|external-player|embedded-mpv).*spec.ts'` — 397 tests passed.
- [x] `pnpm nx run-many -t lint -p portal-stalker-data-access electron-backend` — passed.
- [x] Electron E2E build with the local Homebrew MPV runtime — passed.
- [x] `pnpm exec playwright test --config=apps/electron-backend-e2e/playwright.config.ts apps/electron-backend-e2e/src/stalker-playback-headers.e2e.ts` — both protected video/radio playback tests passed.
- [x] Native seek verification on generated HTTP media and the reported stream, plus CDP/UI verification; no native implementation changes.
- [x] `pnpm run release:notes:validate` and `git diff --check` — passed.

### DIFF
--- a/.changes/stalker-seek-byte-ranges.md
+++ b/.changes/stalker-seek-byte-ranges.md
@@ -1,0 +1,6 @@
+---
+type: fix
+area: stalker
+---
+
+Seeking in Stalker movies and series now reaches the selected position instead of jumping forward or skipping to the next episode after resuming playback.
--- a/apps/electron-backend/src/app/services/stalker-playback-context.service.spec.ts
+++ b/apps/electron-backend/src/app/services/stalker-playback-context.service.spec.ts
@@ -7,6 +7,31 @@ import {
 describe('stalker playback context', () => {
     const macAddress = '00:1A:79:AA:BB:CC';
 
+    it.each([
+        ['http://range-portal.test/c/', 'http://range-cdn.test/episode.mkv'],
+        [
+            'https://range-downgrade.test/c/',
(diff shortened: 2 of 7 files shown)


4gray:
@codex review

PR #1536, opened 2026-09-05
feat(portals): make connection cooldown configurable in desktop settings

4gray:
## What changed

Desktop Settings → General → Portal connections now offers **Pause requests to unavailable portals**, enabled by default. Saving applies the choice immediately to both Xtream and Stalker; it survives restarting the app. Disabling clears active cooldowns, re-enabling starts fresh, and late completions from the previous guard cannot alter the new one. The environment disable switch retains priority. PWA clients do not expose a control for the shared server guard.

Account info distinguishes a skipped request from an unavailable portal, with localized paused-request copy and **Retry now**. Stalker retains cached account data when a refresh is paused. Copy is included in all 19 locales.

## Why

Users investigating intermittent portal failures need a reversible compatibility option without launching from a terminal. Related to #1438, but intentionally does **not** fix or close its separate same-millisecond sibling-counting bug.

Updated `docs/architecture/host-connectivity-guard.md` and mirrored guidance in `AGENTS.md` / `CLAUDE.md`.

## Release note

- [x] Added `.changes/portals-connectivity-setting.md`

## Checks

- [x] Settings unit tests: 18 suites / 173 tests; settings store and runtime capability tests.
- [x] Electron settings IPC, guard, Xtream and Stalker request unit tests: 44 tests.
- [x] Both account-info component suites, including paused errors, real network errors, cached data, and retry recovery.
- [x] Electron E2E: trips a real loopback connection failure, stages the UI edit until Save, bypasses for both protocols, relaunches with the saved opt-out, and re-enables without stale cooldowns. Light/dark setting screenshots inspected.
- [x] `pnpm run typecheck:ci`; lint for all six affected runtime projects plus the added E2E file (no errors; existing project warnings remain).
- [x] `pnpm run i18n:check`, `pnpm run release:notes:validate`, `pnpm run skills:validate`, `git diff --check`.

The local atomized E2E command encountered an occupied mock-server port; the same Playwright spec passed with mock servers on dedicated ports 33210/33211. A supplemental standalone typecheck of the entire Electron E2E tree still reports existing errors in unrelated specs; it reports none in the new spec. Production typechecks and the actual Electron workflow pass.

Issue #1436, opened 2026-08-13
web-backend: re-validate redirect hops in the provider proxy (SSRF)


4gray:
Split out of #1424, where CodeQL alerts #74/#75/#76 (`js/request-forgery`) were dismissed as pre-existing. The alerts are pre-existing — CodeQL is green on master with identical code, and they only surfaced there because that PR touched the `.get(` call lines — but the concern behind them is real and worth closing properly.

## The gap

`apps/web-backend/src/app/web-backend-app.ts` validates a provider URL **once**, at `/provider-targets` registration: scheme, no URL credentials, and (unless `IPTVNATOR_PROXY_ALLOW_PRIVATE_NETWORKS` is set) a DNS resolution whose addresses must not be private or reserved.

After that, the proxy routes call axios with its default transport, which follows up to 21 redirects **without re-validating any of them**. So a registered, validated public URL can `302` to a private or link-local address and the proxy will follow it and return the body:

```
https://public-provider.example/…  →  302  →  http://169.254.169.254/latest/meta-data/
```

Electron does not have this gap: `requestWithValidatedRedirects` (`apps/electron-backend/src/app/util/validated-axios.ts`) uses `maxRedirects: 0` and re-runs the full URL policy, DNS lookup included, on every hop.

Affected routes: `/xtream`, `/stalker`, `/parse`, `/parse-xml` — i.e. all four sites carrying the `// codeql[js/request-forgery]` comment. Those comments are honest about *why* the surface is bounded; they do not claim it is eliminated.

## Why the cheap fix is not enough

axios exposes `beforeRedirect`, which is synchronous. It can re-check the parts that need no I/O — scheme, URL credentials, literal private/reserved IPs — and that alone would stop the cloud-metadata case above.

It does **not** stop DNS rebinding: register a hostname that resolves publicly, pass the hook while the answer is still the public IP, and have the name resolve to `10.x.x.x` on the next lookup. Shipping `beforeRedirect` as "redirect SSRF fixed" would leave that open while recording it as closed, which is worse than the current state where the gap is at least documented.

## Suggested shape

Mirror the Electron implementation rather than inventing a second one:

- A wrapper implementing the existing `WebBackendHttpClient` interface (`get<T>(url, options?)`) that internally does `maxRedirects: 0` and walks the chain itself, re-running `validateProviderUrl` per hop.
[...]

PR #1547, opened 2026-09-05
fix(host-health): retain trial ownership until requests settle

### LINKED ISSUE #1439: host-health: half-open lets a second trial out when one legitimately runs past 45 s
Raised by Codex during the review of #1424, but it belongs to the guard that merged in #1421. Filing it so it is not lost. Scoped carefully: the worst consequence was **already fixed** by the `trialId` work in that PR — what is left is narrower than the original report.

## What happens

Half-open lets exactly one request through and holds the rest back until it settles. The slot is released either when that request reports, or by elapsed time:

```ts
const trialStale =
    state.trialStartedAt !== null &&
    now - state.trialStartedAt >= TRIAL_TIMEOUT_MS;   // 45_000
if (state.trialStartedAt !== null && !trialStale) {
    return { allowed: false, retryAfterMs: 0 };
}
```

`TRIAL_TIMEOUT_MS` is documented as a leak backstop:

> Above the longest request timeout (30 s) plus margin, so it only fires if a caller leaked the token.

That premise does not hold, and the file says so a few lines earlier, in the `trialId` docblock:

> A trial can outlive its own window — `requestWithValidatedRedirects` gives each of up to five redirect hops its own 30 s budget

There is a second route to the same place: axios' `timeout` on the default transport is not a wall-clock deadline. It bounds time-to-headers and then continues as the socket's *inactivity* timeout, so a slow-but-alive transfer can legitimately run well past 45 s without ever idling 30 s.

Either way a healthy trial is declared stale, and a second request is admitted while the first is still in flight.

## What is already handled

`trialId` means the abandoned trial can no longer free the replacement's slot or corrupt its state — `ownsTrial` gates that, and its late failure is deliberately treated as ordinary evidence. So this is not a state-corruption bug.

## What remains

The one-request guarantee itself. During half-open against a slow endpoint, more than one request goes out, which is the thing half-open exists to prevent — the guard's own words are "not a whole screenful". The practical cost is small (an extra request or two, not a fan-out), which is why this is low priority rather than urgent.

## Suggested fix

Release the slot on *liveness* rather than elapsed wall time. Options, roughly in order of intrusiveness:

1. Have the caller signal that a trial is still in flight, so the backstop only fires when nothing has reported — this is the honest reading of "it only fires if a caller leaked the token".
2. Keep the timer but size it from the real worst case rather than a single request timeout: 5 hops × 30 s for `requestWithValidatedRedirects`, and no fixed bound at all for an idle-based body transfer. A bigger constant is a weaker fix, since the idle case has no ceiling.

Option 1 is the one that actually matches the documented intent.

## Related

`docs/architecture/host-connectivity-guard.md` describes the timeout as idle-based in the "Request timeouts" section; the 45 s constant predates that note and was never reconciled with it.

4gray:
## Summary

A half-open portal probe could legitimately remain active past the guard's 45-second expiry, either across separately timed redirect hops or while receiving an active response body. The expiry admitted a second probe before the first settled.

Retain the trial slot for the complete request lifetime. All four Electron/PWA portal request owners release in `finally`, independently of outcome reporting, so cancellations and exceptions in diagnostics or reporting cannot strand the slot. Cleanup also releases while the environment kill switch is enabled. No heartbeat timer or larger deadline is introduced.

The existing trial ID/epoch protection, monotonic admission ordering and concurrent-failure deduplication (#1537), desktop preference invalidation (#1536), and redirect failure attribution remain intact. Explicit resets, reachability evidence and bounded state eviction keep their existing semantics. Redirect SSRF #1436 is outside this change.

## Validation

**Ready to merge:** all 20 GitHub checks are green on `0011878b09fc46a2d6722977992af79e0cbbf23d`, including unit/typechecks, lint, web E2E, Electron E2E on Ubuntu/macOS/Windows, Docker and every platform package build. GitHub reports `CLEAN` / `MERGEABLE`; there are no review threads. The long runner queue required retrying only the three unstarted jobs; successful jobs were preserved and no code changes were needed. The PR is intentionally left unmerged.

- Regression proof against the original implementation: 4 shared-guard, 7 Electron and 5 web-backend tests fail at the intended assertions; the fix was restored afterwards.
- `pnpm nx run-many -t test -p shared-host-health,web-backend --parallel=2 --runInBand` — 52 shared guard + 66 backend tests pass.
- `pnpm nx test electron-backend --runInBand '--testPathPatterns=(xtream.events|stalker.events|host-connectivity-guard|validated-axios).spec.ts'` — 59 tests pass.
- Coverage includes pending probes beyond 45 seconds and the idle TTL, real axios redirect + unfinished streaming response, Electron multi-hop redirects, success/failure/cancellation, late results after reset, cleanup after reporting throws, and environment-disable settlement.
- Lint passes for all three affected projects (existing warnings only); shared-host-health and web-backend builds pass, and the Electron E2E build completed.
- `pnpm exec playwright test --config apps/electron-backend-e2e/playwright.config.ts settings-connectivity-guard.e2e.ts` — 1 passed (real Electron IPC/network failures, saved preference and restart). The atomized Nx launch initially failed while starting mock servers; the direct run passed against the completed E2E build. CI additionally covers the affected workflows across platforms.
- `pnpm run release:notes:validate` and `git diff --check` pass. Updated the canonical host-connectivity contract, mirrored AGENTS/CLAUDE guidance and added `.changes/host-health-live-trial.md` for the next release (0.24).
- Independent local code review: no actionable findings. GitHub reviews on `0011878b09fc46a2d6722977992af79e0cbbf23d`: Greptile **5/5**, Codex **no major issues**, no inline review findings.

## Lifecycle boundary

Cancellation releases after the outbound transport settles, not merely when abort is requested. Stalker IPC and a disconnected PWA client do not currently cancel their outbound transport. Existing transport timeouts handle inactivity; a custom transport that never settles must supply cancellation at its own layer. Future owners must preserve the documented `try`/`finally` contract.

Closes #1439

### DIFF
--- a/.changes/host-health-live-trial.md
+++ b/.changes/host-health-live-trial.md
@@ -1,0 +1,7 @@
+---
+type: fix
+area: host-health
+issues: [1439]
+---
+
+Portal recovery now waits for an active probe to finish before sending another request, even when redirects or a slow response take longer than 45 seconds. Completed and cancelled requests release the probe slot reliably.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -176,6 +176,12 @@ categories by provider ID and type. See `docs/architecture/category-management.m
 
 ## Portal Connectivity Preference
 
+- Half-open trial slots follow the complete request lifetime with no elapsed-time
+  expiry. All four Electron/web-backend portal handlers release in `finally`,
+  independently of outcome reporting; cleanup preserves trial/epoch ownership
+  and works while the environment override is disabled. Contract:
+  `docs/architecture/host-connectivity-guard.md` ("Trial ownership follows the
+  request lifetime").
 - Desktop Settings > General > Portal connections exposes default-on
   `Settings.portalConnectivityGuard`. Only explicit false opts out. Save mirrors
   the value to Electron `PORTAL_CONNECTIVITY_GUARD` and applies it without restart;
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1661,6 +1661,12 @@ No formal migration system yet. Schema changes are applied via raw SQL in the `c
 
 ## Portal Connectivity Preference
 
+- Half-open trial slots follow the complete request lifetime with no elapsed-time
+  expiry. All four Electron/web-backend portal handlers release in `finally`,
+  independently of outcome reporting; cleanup preserves trial/epoch ownership
+  and works while the environment override is disabled. Contract:
+  `docs/architecture/host-connectivity-guard.md` ("Trial ownership follows the
+  request lifetime").
 - Desktop Settings > General > Portal connections exposes default-on
   `Settings.portalConnectivityGuard`. Only explicit false opts out. Save mirrors
   the value to Electron `PORTAL_CONNECTIVITY_GUARD` and applies it without restart;
--- a/apps/electron-backend/src/app/events/stalker.events.spec.ts
+++ b/apps/electron-backend/src/app/events/stalker.events.spec.ts
@@ -83,6 +83,75 @@ describe('StalkerEvents host connectivity guard', () => {
         consoleWarnSpy.mockRestore();
     });
 
+    it('holds a live trial beyond 45 seconds and releases a cancelled request', async () => {
+        let now = 1_000;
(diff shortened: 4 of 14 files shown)


4gray:
@codex review

PR #1610, opened 2026-09-15
fix(epg): re-pick the Xtream channel list's current program once it ends

### LINKED ISSUE #767: EPG does not update in channel list
**Describe the bug**
In Live TV, when entering a category, a channel list is presented on the left side. Under the channel name, the current program EPG data is shown. After the end time of that program, the EPG shown is not updated to what is currently showing.

**To Reproduce**

1. Start watching a live tv channel.
2. Continue watch it until after the end time of current program.
3. Look at EPG in channel list; it is still showing the previous program.

**Expected behavior**
Current program in channel list should be updated with the actual current program.

**Screenshots**
Image

In screenshot, "OS 2026: Huvudsandning" is not the current program: "OS 2026: Ishockey" is. In the screenshot its only 1 minute past but the bug is valid even long time after.

- Windows  version 0.18.0
- Xtream code playlist

**Additional context**
Backing out from view by pressing "< Channels" in top left, and then entering the same category again will update the list properly. 

The program list under "Electronic Program Guide" does not update automatically either, but a switch in channel will do that.

thejdubb02:
## What changed
Xtream Live TV's channel list now re-checks each visible channel's shown "current program" on a 60 second timer, so it advances once that program ends. loadEpgForVisibleChannels() (portal-channels-list.component.ts) gains an optional `force` flag that lets the timer re-apply a warm cache or trigger a real re-fetch for a channel that is already tracked, something it previously skipped. The map is never cleared, so a row keeps showing its last known program instead of going blank while a refetch is in flight.

## Why
epgPrograms (line 164) is written only by applyProgram() (line 532), which only runs on scroll-into-view, a new EPG result arriving, or an EPG display-offset change. Nothing re-evaluated it as wall-clock time passed, so once a program's end time arrived the row stayed pinned on it until the category was left and re-entered (#767), which forces a full remount. Mirrors the EPG refresh interval already used by the M3U channel-list-container sibling.

## Release note
- [x] Added a note under `.changes/`

## Checks
- [x] `pnpm run lint` and the affected `pnpm nx test <project>` pass

This PR was drafted with AI assistance (Claude) under human review; the diagnosis, fix, and tests were verified against the codebase and the project's own test suite before submitting.

Fixes #767


4gray:
Thanks for this, @thejdubb02 — the diagnosis is exactly right, and it is the part that was hard. Nothing re-evaluated the row as wall-clock time passed, and the tick plus the three tests you wrote are a good shape for the fix. Two things came up while reviewing, so I've taken it forward rather than leaving it in your queue.

**The row can move backwards.** `pickPreviewProgram()` falls back to the earliest known item when nothing is on air, and the forced re-apply hands that back to the row every minute. With a cached guide of `A 05:30–06:00` and `B 06:00–06:02`, the row correctly reaches B at 06:01, then at 06:03 both have ended and it jumps back to A — and stays there until the five-minute cache expires, with no refetch, because the queue skips any stream that still holds a cached answer.

That is not a corner case. The uploaded-XMLTV fallback caches exactly one programme, the current one, so every channel using it lands in that state the moment its programme ends. The reporter's screenshot in #767 is a minute past the hour, which is the window this leaves stale.

**The forced reload also costs traffic it does not need to.** Every visible row was re-requested once its cache aged out, whether or not its programme had ended — roughly 30 requests a minute for a full viewport against the queue that exists to keep providers from banning us.

#1647 keeps your commit with your authorship and your three tests unchanged, and makes the tick a staleness check instead: a programme still on air only has its progress bar advanced (those bars never moved before, which turned out to be a second half of the same bug), a finished one is replaced only by something on air or upcoming, an exhausted cache is dropped and refetched at most once per cache lifetime, and a channel the provider has no guide for is left alone. The row still never blanks out while a refill is in flight, which was your design.

Closing this in favour of #1647 — review very welcome there, and thanks again for digging into it.

4gray:
This shipped — merged as 42b41ebab, closing #767, with your commit and your authorship in the history and your co-authorship on the squash. Thanks again, @thejdubb02: the diagnosis was the hard part and it was right.

For what it is worth, the review that followed found eight more things on top of the one I described here, mostly around the fact that a live layout mounts that channel list more than once over a single EPG queue. The final shape is documented in `docs/architecture/m3u-playlist-module.md` under "Xtream channel-row programme refresh" if you are curious.

PR #1612, opened 2026-09-15
fix(search): make title search case folding locale-invariant (#609)

### LINKED ISSUE #609: search fails for some upper/lower case characters
I see search fails when i use lowercase/uppercase for some characters. This happens for "ç/Ç" "i/İ" "Ş/ş" and "ü/Ü".

1. Make a search for:
"İnş" and "inş"
"Ünlü" and "ünlü"
"Çan" and "çan"
"Şan" and "şan"
 
2. Compare results.

**Expected behavior**
Search results should be same for lower and uppercase chars.

Happens on beta5 on both Windows and GNU/Linux.

thejdubb02:
## What changed

Search now returns the same results for the upper and lower case forms of
Turkish characters, so "inş" and "İnş" (and "Çan"/"çan", "Şan"/"şan",
"Ünlü"/"ünlü") match the same channels and movies. The content search
normalizer now folds case with toLowerCase / toUpperCase instead of the
locale-sensitive toLocaleLowerCase / toLocaleUpperCase, and strips combining
marks in the SQL token path before it splits the text into words.

## Why

Fixes #609.

Case folding in content-search.util.ts used toLocaleLowerCase, which depends
on the OS locale, and that broke Turkish search two ways:

1. Folding the dotted capital I ("İ", U+0130) leaves "i" plus a combining dot
   above (U+0307). normalizeSqlSearchText treated that mark as a word
   separator, so "İnş" became the two tokens "i" and "nş" while "inş" stayed a
   single token. The one-character token routed "İnş" to the short-token
   prefix index (anchored to the start of the title) while "inş" used the
   trigram FTS path, so the two searched differently.
2. Under a Turkish or Azeri OS locale, toLocaleLowerCase maps "İ" and ASCII
   "I" to different letters than "i", so scoreSearchTextMatch dropped the same
   title for the lower case query, and the query no longer lined up with the
   SQLite FTS index, which folds locale-invariantly.

Folding with toLowerCase / toUpperCase removes the locale dependence (and
matches how the FTS index folds), and stripping the U+0300 to U+036F combining
marks in normalizeSqlSearchText keeps the dotted I from fracturing (the same
range normalizeSearchMatchText already removes). Precomposed letters such as
ç, ş, ü and é are outside that range, so accent handling is unchanged.

Scope note: this covers the database-backed content search used by global
search and M3U payload search. One renderer-side in-memory filter,
filterBySearchTerm in libs/portal/xtream/data-access (with-selection.feature.ts),
still does a naive title.toLocaleLowerCase().includes(...) and has the same
dotted-I limitation. I left it out to keep this focused, happy to follow up on
it if you want it in the same change.

## Release note

- [x] Added a note under `.changes/` (`.changes/search-turkish-case-folding.md`)
- [ ] Not needed

## Checks

- [x] `pnpm run lint` and `pnpm nx test electron-backend` pass (typecheck too)

AI assistance: I used AI assistance to research, write and test this change,
then reviewed and verified it myself before submitting.


4gray:
Thanks a lot for this, @thejdubb02 — the root-cause analysis was spot on, and the combining-dot split into "i" + "nş" was exactly what sent "İnş" to the wrong index.

While verifying it against a real catalog I found that the fix reached only the trigram FTS arm of the global search. Two other paths still missed "İnşaat" for a lower-case "inş":

- the LIKE/GLOB arms (per-playlist Xtream search, content scan, the M3U payload prefilter), because `"inş".toUpperCase()` is `INŞ`, so no pattern ever spells the dotted İ;
- every in-memory list filter in the renderer (M3U sidebar, Xtream channel list and category filter, Stalker, collections, …), which all do `title.toLowerCase().includes(term)` — and `"İnşaat".toLowerCase()` still carries the combining dot, so `.includes("inş")` is `false`.

Your PR doesn't allow maintainer edits, so I opened #1640 with your commit kept as-is (rebased on master) plus a second commit that adds explicit `'tr'`-locale İ variants to the pattern builders and a shared `foldSearchText()` for the renderer filters, with a SQLite-backed spec over the real trigram index. I'll close this one in favour of #1640 — your authorship stays on the commit. Thanks again!

4gray:
Superseded by #1640, which keeps this commit and completes the remaining search paths.

4gray:
For the record: the completion of this work merged as b30c783e8 (#1640), with your commit kept and your authorship credited on the squash. Thanks again @thejdubb02 — issue #609 is closed.

omeringen:
@thejdubb02 thanks for the fix.

> The new fold still misses the Turkish I/ı pair.

Gonna open separate issue about it.

PR #1613, opened 2026-09-15
fix(shell): persist and restore the app zoom level (#1109)

### LINKED ISSUE #1109: Zoom level not persisted across restarts or section changes
## Description

The zoom level (set via cmd+`+` and cmd+`-`) is not properly saved and restored. This makes IPTVnator unusable for visually impaired users who need a larger font size.

## Bug — Zoom resets on section change

1. Set a custom zoom level
2. Close and reopen IPTVnator
3. The correct zoom level appears briefly on the last-viewed section or sometimes is not persisted at all
4. Switch to any other section (e.g. from Favourites to All Channels)
5. → Zoom resets to default

This inconsistency (zoom correct on first view, then reset on navigation) suggests the value is stored but not applied consistently across all components.

## Bug — Window position not remembered (regression)

#205 was reported and marked as fixed in v0.17.1, but the behaviour has regressed in v0.21.0: the window always reopens at the default position and size, not where it was last placed.

## Feature requests

- **Persist zoom level** across restarts and section changes
- **Persist window position and size** across restarts (see also: #205)
- **Option to always open maximized** (useful for TV/living room setups and visually impaired users)

## Environment

- OS: macOS
- IPTVnator version: 0.21.0

thejdubb02:
## What changed

The app-wide zoom level (Cmd/Ctrl and +/-) is now saved and restored across restarts. A `did-finish-load` listener reapplies the stored zoom on every renderer load, and the level is saved next to the window bounds when the window closes and before the app quits. A small `persistWindowState` helper replaces the two duplicated bounds-only saves.

The zoom is saved only from a window that has finished loading, so a close during startup cannot overwrite the stored level with the default. A non-finite stored value is ignored, and a destroyed webContents is never read.

## Why

Issue #1109 reports that the zoom level does not persist, which makes the app hard to use for anyone who relies on a larger font size. Nothing saved or restored the zoom, so it reset to the default on every launch. Because the restore runs on `did-finish-load`, it also covers a full renderer reload, and in-app section changes are SPA routing that never reloads, so the level holds between them.

The issue lists two further requests, persisting window position and size and an option to always open maximized. Both are already implemented on current master (the `WINDOW_BOUNDS` restore in `initMainWindow`, and the "Window on startup" setting), so this change covers the remaining zoom part.

## Release note

- [x] Added a note under `.changes/` (`shell-zoom-level-persistence.md`)
- [ ] Not needed

## Checks

- [x] Tests added for the changed behavior: eight cases in `app.spec.ts` (restore on load, reapply on reload, no-op when unset, ignore a non-finite value, save on close, no clobber when closed before load, skip a destroyed webContents, save on before-quit). The behavioral cases were confirmed to fail on master before the fix.
- [x] `pnpm nx test electron-backend --coverage` (180 suites, 2273 tests), `pnpm nx lint electron-backend` (0 errors), `pnpm run typecheck:backend`, `pnpm run i18n:check`, and `pnpm run release:notes:validate` all pass.

This change was written with AI assistance and reviewed by a human before submission.

Fixes #1109


4gray:
Thanks a lot for taking this on, @thejdubb02 — the persistence half of this (the store key, the shared `persistWindowState` helper, the close/quit save, the note) is exactly right and is kept as your commit in the follow-up.

I could not push to this branch (maintainer edits are off), so I finished it in #1617 on top of your commit. What I had to change, and why: the restore only works in dev mode. The packaged app runs under `file://` with path routing, and Chromium keys zoom by the full URL there, so `webContents.setZoomLevel` on `did-finish-load` sets the level for `index.html` only — the first window resize after switching sections snapped it back to the default, and the close handler then saved that route's 0 over the user's level. Restoring through `webFrame.setZoomLevel` in the preload (a temporary, frame-bound zoom) fixes that, and saving on `did-start-navigation` closes the reload gap Greptile flagged. There is a packaged-mode E2E in the new PR that fails on this branch at that exact step, if you want to see it.

Closing this one in favour of #1617; you are credited there as the author of the first commit.

PR #1617, opened 2026-09-17
fix(shell): persist the app zoom level as frame-bound temporary zoom (#1109)

### LINKED ISSUE #1109: Zoom level not persisted across restarts or section changes

4gray:
Supersedes #1613 by @thejdubb02 — their commit is kept as the first commit of this branch (the `ZOOM_LEVEL` store key, the shared `persistWindowState` helper, the release note and the close/quit save path all come from it). Thank you!

## What changed

The app-wide zoom level (Cmd/Ctrl and +/−) is restored on every document load and saved back on close, quit and before a reload, so it survives section changes, window resizes and restarts.

The restore moved from the main process into the preload: `applyPersistedZoomLevel` asks for the stored level over a new synchronous `WINDOW:GET_ZOOM_LEVEL` IPC and applies it with `webFrame.setZoomLevel`. The main-process side (`services/window-zoom-level.ts`) answers that IPC, marks the sender as owning the level, and writes the live level to electron-conf from `close`, `before-quit` and every main-frame cross-document `did-start-navigation`.

## Why

#1613 applied the level with `webContents.setZoomLevel` on `did-finish-load`, which works in dev mode but not in the packaged app. The packaged renderer runs under `file://` with path routing, and Chromium keys zoom by the **full URL** when a URL has no host — so every `pushState` to another section owns its own zoom entry. Verified with a standalone Electron 41 script and with the new E2E: after a section change the next window resize snapped the renderer back to the default (exactly the "appears briefly, then resets" from #1109), and the close handler then saved the current route's entry (usually 0) over the user's level. Chromium already persists those per-URL entries on its own, which is why the level used to flash on `index.html` at startup.

`webFrame.setZoomLevel` installs a *temporary*, frame-bound zoom level instead: it survives in-page navigation and resizes, the macOS menu roles increment it, and `getZoomLevel()` reports it regardless of the route — so the save path from #1613 becomes correct. Chromium never persists temporary zoom, so the electron-conf write stays necessary; saving on `did-start-navigation` also closes the "reload restores a stale level" gap Greptile raised, since a reload drops the temporary level before the new document's preload reads the store.

Windows/Linux have no zoom shortcuts today (`setMenu(null)`, no `zoom-changed` handler), so in practice this fixes macOS; the persistence is platform-neutral and will pick up any future shortcut routed through `webContents.setZoomLevel`.

Contract: `docs/architecture/workspace-shell.md`, "Zoom level".

## Release note

- [x] Added a note under `.changes/` (`shell-zoom-level-persistence.md`, reworded)
- [ ] Not needed

## Checks

- [x] Tests added or updated: `window-zoom-level.spec.ts` (read/persist/ownership/navigation gating), `preload-zoom-level.spec.ts`, `main.preload.spec.ts` (synchronous restore before the bridge is exposed), `window.events.spec.ts` (sync IPC), `app.spec.ts` (zoom block rewritten for the new contract), and a new Electron E2E `window-zoom-level.e2e.ts` measuring the rendered zoom factor across a section change, a resize, a cross-document reload and a restart — it fails on the #1613 approach at the resize-after-section-change step (factor 1 instead of 1.2) and passes here.
- [x] `pnpm nx test electron-backend` (182 suites, 2293 tests), `pnpm nx run-many -t lint -p electron-backend electron-backend-e2e shared-interfaces`, `pnpm run typecheck:ci`, `pnpm run release:notes:validate`, and the new E2E via Playwright directly (the nx target's mock-server dependencies collide with another worktree's ports locally).

Closes #1109


4gray:
Follow-up pushed. The Ubuntu/Windows Electron E2E failure (`legacy-playlist-migration.e2e.ts` › defer-epg, `#initial-splash` still present) was a real regression of the first commit, not a flake: a `webFrame.setZoomLevel` at preload start leaves a hidden window without a first frame on Linux/Windows — `ready-to-show` never fires, `show()` never runs, no animation frames, so the splash that `main.ts` removes in a `requestAnimationFrame` stays. Reproduced in a Linux container (Electron 41 under xvfb): preload-start and `setTimeout(0)` both break it, `DOMContentLoaded`, `load` and a main-process push after `did-finish-load` all keep `ready-to-show` and the zoom still sticks across pushState + resize. macOS is unaffected and CDP-driven tests force frames, which is why only that one assertion caught it.

The preload now applies the level at `DOMContentLoaded` and acknowledges with `WINDOW:ZOOM_LEVEL_APPLIED`; ownership for the save path follows the acknowledgement instead of the request. Greptile's release-note point is addressed too (macOS-only shortcut wording). Locally: the affected unit specs, lint, `typecheck:ci`, and the zoom + legacy-migration E2E files on macOS are green.

4gray:
@greptileai review

PR #1620, opened 2026-09-18
feat(playback): fullscreen episode panel for series playback

4gray:
## Summary

Series playing in fullscreen now get the same slide-in side panel the live channel list has — with **season tabs and the episode list** instead of channels. Rest the mouse on the left edge, click it, or press `C`; pick an episode and it plays inline **without leaving fullscreen**.

> Builds on #1619 (the panel discoverability fix): its commit is included here, so that part of the diff duplicates #1619 until it lands, after which a rebase onto `master` drops it. Targeting `master` directly because `ci.yml` only runs for PRs against `master`.

### Panel generalization (`libs/ui/playback/src/lib/fullscreen-channel-panel/`)
- `FullscreenChannelPanelHost` gains optional `panelSearchEnabled` (false drops the search field for a title row; `C` then focuses the panel itself) and `panelKind: 'channels' | 'episodes'` (accessible names and the close button's label only).
- The template context gains `open: Signal<boolean>`, so a body that stays mounted between openings can react to the panel coming up.
- Every pointer/keyboard rule is untouched; the four live providers (M3U, Xtream, Stalker, unified live tab) and their specs are unchanged.

### Series host (`PortalInlinePlayerComponent`)
- Provides `FULLSCREEN_CHANNEL_PANEL` for the nested `app-web-player-view` — the same component that feeds the Up Next rail, so an episode click travels `upNextEpisodeSelected` and reuses the exact inline play path (fullscreen survives the engine remount). Being the nearest provider it also shields the nested view from the M3U page-level channel provider.
- New `app-fullscreen-episode-panel` (`libs/ui/playback/src/lib/fullscreen-episode-panel/`): `SeasonTabsComponent` on top (now exported from `@iptvnator/ui/components`), then the selected season's rows — TMDB still or a large numeral tile when there is none (or the image fails), `S01E03` label, runtime, 3-line clamped overview, watch-progress bar, watched check, now-playing marker. The tab follows the playing episode's season (`linkedSignal`) and the playing row is centred on open via `scrollTop` math (effect keyed on primitives, so progress ticks never yank a scrolled list).
- Hosts pass `seriesEpisodes` / `episodePlaybackPositions` / `seasonLoadStates` and wire `episodePanelSeasonSelected` to their existing `onSeasonSelected`, so Xtream's TMDB season enrichment and Stalker's lazy VOD season load run for the panel's season (a spinner row while in flight, a Retry row after a failed request). The host object is built by `createEpisodePanelHost()` in its own file, keeping the inline player component under the size guideline.
- Gates: `Settings.fullscreenChannelPanel` (label now "Channel and episode list in fullscreen", reworded in all 19 locales), `contentType === 'episode'` (movies never get it), native-view Embedded MPV withheld by the view, external MPV/VLC never mount the inline player.

### Visual check
Captured against the Xtream mock in Chromium (1440×810): rows with TMDB-style stills, the numeral-tile fallback when every still is blocked, and season 2 with the "Back to playing episode" chip. The season tabs pin the light-on-dark tokens, since the app's light theme resolves them to dark ink on the dark panel.

## Test plan
- [x] `ui-playback` unit suite: 1442 → 1465 tests green (new specs: `fullscreen-episode-panel.util.spec.ts`, `fullscreen-episode-panel.component.spec.ts`, `portal-inline-player-episode-panel.spec.ts`; `fullscreen-channel-panel.component.spec.ts` extended with a search-less `episodes` host)
- [x] `portal-xtream-feature` serial-details specs (72) and `portal-stalker-feature` series-view specs (152) green with the updated inline-player stubs
- [x] `pnpm run typecheck:web`, `pnpm nx build web` (template type-check), lint on `ui-playback`, `components`, `portal-xtream-feature`, `portal-stalker-feature`, `web-e2e`
- [x] E2E: `xtream-series-playback.e2e.ts` (split out of `xtream.e2e.ts`, which had hit the 1200-line spec cap) — 3/3 Chromium, including the new fullscreen episode switch through the panel across two seasons
- [x] `pnpm run release:notes:validate`, `pnpm run i18n:check`
- [ ] Electron: native-view Embedded MPV still withholds the panel (gate unchanged, not re-verified live)


4gray:
@greptileai — both findings addressed in the follow-up commit: failed lazy seasons now show a Retry row instead of an endless spinner (per-season load state), and episode rows are plain buttons inside list items. Also rebased onto the #1619 branch so the PR is no longer conflicting.

4gray:
@greptileai — size finding addressed: the panel host now lives in `portal-inline-player-episode-panel.host.ts` (`createEpisodePanelHost()`), and the Stalker season load state moved to a pure helper in data-access. PR retargeted to master so CI runs.

4gray:
@greptileai — rebased onto master now that #1619 has landed; the branch carries only the episode panel commits.

PR #1637, opened 2026-09-19
fix(dashboard): live rails find XMLTV programmes outside the global EPG sources

4gray:
## Problem

The dashboard's "Now on air on favorite channels" and "Recently watched Live TV" rails showed only the LIVE chip for channels that have a current programme on the "See all" pages.

Root cause: the rails call the batched XMLTV lookup with no playlist scope, so `EpgService` scoped it to the Settings-managed global EPG URLs **only**, with no fallback. The "See all" pages resolve through `StreamResolverService`, which never scopes by source. A favourite whose guide lives in another playlist's XMLTV (or whose programmes are tagged with a source other than its channel row — the multi-import case) had a programme in the list and none on the dashboard. Reproduced on a real profile.

## Fix

- `EpgLookupOptions.anySourceFallback` (renderer-only, never forwarded to the bridge): after the scope — playlist sources, then global — has answered, the keys still without a programme are retried once through the existing source-less batch path and its cache. `getCurrentProgramsForChannels` was split into the scoped decision, the any-source merge and the extracted unscoped lookup; behaviour without the option is unchanged.
- The dashboard rails pass the option. The channel list keeps the strict scope (the existing "does not fall back to the unscoped EPG pool" spec still holds).

Not covered here, by design: Xtream/Stalker cards. They carry no XMLTV key and their EPG comes from the portal. Follow-up PR (plan below) adds portal EPG to the rails, loaded lazily per visible card.

## Tests

- `epg.service.spec.ts`: 4 new cases — global-scope miss retried unscoped with only the unresolved keys and no `sourceUrls`; scoped null kept when the pool has nothing; no retry when everything resolved in scope and never without the option; playlist scope → global → any-source order.
- `epg.e2e.ts` (Electron): playlist-declared XMLTV + a global XMLTV that lacks the channel → the favourite's dashboard card shows the programme. Verified to fail on the old code (`Received: " M3U · Live "`).
- `pnpm nx test epg-data-access` 50/50, `workspace-dashboard-feature` 64/64, lint both projects + e2e, `pnpm run typecheck:ci`, full `epg.e2e.ts` 13/13, `release:notes:validate`.

## Docs

- `.changes/dashboard-live-rails-epg-any-source.md` (fix).
- `docs/architecture/m3u-playlist-module.md` — the "Scoped lookups" bullet documents the opt-out and who uses it; `CLAUDE.md` EPG section mirrors it in one line.

## Follow-up plan: portal EPG on the dashboard, per-card lazy

Separate PR (feature, network cost profile):
- `DashboardLiveEpgService`: per-card signal (`pending` / programme / none), 60 s TTL + 30 s failure cooldown, queue with max 2 in flight and 200 ms spacing (same numbers `EpgQueueService` already uses); Xtream via `EpgQueueService`, Stalker via `get_short_epg` with `shortEpgWindowSize` — the per-channel fetch extracted from `StreamResolverService` into a shared helper so the two surfaces cannot drift; manual mappings first; display-offset tagging like `EpgService`.
- Visibility-driven: `lib-dashboard-rail` reports visible cards through an `IntersectionObserver`; only visible cards enter the queue, each card fills its own row when its answer lands, the page never waits.
- The 30 s tick only recomputes progress from cache; refetch when the programme ended or TTL expired and the card is visible.
- XMLTV/M3U cards keep this PR's single batch (one local SQLite query, already non-blocking).
- Tests: queue bounds, TTL/cooldown, visibility subscribe/unsubscribe, incremental updates; Electron E2E against the Xtream and Stalker mocks asserting hidden cards are never requested.


4gray:
@codex review

4gray:
Both findings fixed in 104c6e01c.

**Greptile P1 — fallback loses playlist identity.** Valid: one flat, source-less map keyed by lookup key alone could hand a card the programme another playlist's guide resolved for the same XMLTV id. The dashboard now groups its lookup by the source scope each card's playlist declares (the channel list's rule: only an M3U playlist carries its own guide; portals use the Settings URLs) and namespaces the answers by that scope (`liveEpgScopeKey` / `liveEpgProgramKey`), so a card only ever reads its own. Playlists pointing at the same guide still share one lookup, and every group keeps the any-source retry.

**Codex P2 — any-source retry on legacy preloads.** Valid: the per-channel path answered through `getCurrentProgramForChannel`, which re-applies the Settings scope, so the retry re-asked the question the scoped pass had already answered. It now uses an unscoped per-channel lookup; every other caller keeps the historical behaviour there.

The regression E2E was rewritten in the process. The old fixture gave the favourite's playlist its own guide, so the card resolved in its own scope and the test would have passed without the fallback. It now matches the reported case — the favourite's playlist declares no guide, Settings hold a global XMLTV that does not know the channel, and the programme exists only in the guide another playlist imported — and was verified to fail with the fallback disabled.

New coverage: per-scope grouping and shared-guide deduplication, the cross-playlist collision, scope-key normalisation, and both legacy-preload paths. `epg-data-access` 52/52, `workspace-dashboard-feature` 67/67, `epg.e2e.ts` 13/13, lint + `typecheck:ci` clean.

4gray:
@codex review

4gray:
@greptileai

4gray:
@codex review

4gray:
@codex review

4gray:
@greptileai

4gray:
Both follow-ups fixed in 2ef0afbec.

**Greptile — oversized services.** Valid, and repo policy I had ignored: this PR grew two files that were already past the limit. Extracted, with no behaviour change:

- `EpgProgramCache` — the 60 s "currently airing" memory, the in-flight request slots and the scope/offset cache keys.
- `EpgChannelMetadataLookup` — icons and display names, same playlist-first / Settings-fallback ladder.
- `DashboardLiveEpgPresenter` — the rails' XMLTV lookup and card enrichment, component-provided so the 30 s heartbeat dies with the page.

`epg.service.ts` 901 → 727, the rails component 1000 → 842. Both are now smaller than before this PR, the three new files are well under 400 lines, and `generate-max-lines-baseline.mjs` produces an unchanged baseline.

**Codex P2 — any-source fallback on portal cards.** Valid. An Xtream or Stalker card carries no XMLTV key, so its lookup key is just its display title; widening that to every imported guide would let a same-named M3U channel answer for a portal channel. The retry is now gated on the card actually having a key, and the flag is part of the scope identity so a guide-less M3U card and a portal card can never share an answer. Portal cards keep the strict scope — their own programmes come from the portal, which is what #1638 adds.

`epg-data-access` 53/53, `workspace-dashboard-feature` 68/68 (new: portal-vs-M3U grouping, scope-key identity), `epg.e2e.ts` 13/13, lint + `typecheck:ci` clean.

4gray:
@codex review

4gray:
@greptileai

4gray:
@codex review

4gray:
@greptileai

4gray:
Both remaining findings closed in 9d879ad0a and 3a7fb9d4b.

**Source file appears binary.** Real defect, mine: the scope and programme keys used U+0000/U+0001/U+0002 as separators and I had written them as actual control characters, so the file was classified as binary and textual diffs stopped showing. The keys are JSON now, which is unambiguous for arbitrary URLs and channel titles alike and keeps the file text.

**Oversized services need extraction.** Finished rather than argued: `epg.service.ts` is **247 lines** and has **left `max-lines-baseline.mjs`** — the baseline the repo says must only shrink is one file shorter than before this PR. The lookup machinery moved out unchanged into `EpgSingleProgramLookup` (one channel and the ladder it walks), `EpgCurrentProgramsLookup` (the batch ladder, backed by the per-channel lookups so both shapes answer identically), `EpgLookupContext` and two pure normalizers. The service keeps the import pipeline, the capability guards and two delegating entry points.

Sizes now, against master: `epg.service.ts` 769 → 247, `workspace-dashboard-rails.component.ts` 886 → 842. Every new file is under the limit; `generate-max-lines-baseline.mjs` removes one entry and adds none.

`epg-data-access` 53/53, `workspace-dashboard-feature` 72/72, `web` EPG specs 10/10, `epg.e2e.ts` 13/13, lint and `typecheck:ci` clean.

4gray:
@greptileai

4gray:
@codex review

4gray:
Split again in c5282e328: `EpgScopedBatchLookup` now owns how ONE scope is asked (the per-channel cache in front of the request, the order-insensitive dedup of identical in-flight batches, the fallback-scope retry), and `EpgCurrentProgramsLookup` owns which scope to ask plus the any-source retry.

Raw line counts in the lib, all inside the limit: service 249, current-programs 256, scoped-batch 180, single-program 129, cache 166, channel-metadata 101. `epg.service.ts` remains out of the max-lines baseline.

`epg-data-access` 53/53, `epg.e2e.ts` 13/13, lint and `typecheck:ci` clean.

4gray:
@greptileai

PR #1638, opened 2026-09-19
feat(dashboard): portal EPG on the live rails, loaded lazily per visible card

4gray:
## What

The dashboard's "Now on air on favorite channels" and "Recently watched Live TV" rails showed only the LIVE chip for Xtream and Stalker channels: they have no XMLTV key, and the rails never asked the portal. Follow-up to #1637 (which fixed the XMLTV scope for M3U cards).

Portal cards now get their programme from the portal, **lazily and per card**:

- `buildDashboardPortalLiveEpgEntry` (dashboard data-access) turns a live `PortalActivityItem` into the `UnifiedCollectionItem` the "See all" pages hand `StreamResolverService.loadEpgForItems`, keyed by the collection uid, so favourites and recent rows of one channel share the answer. Radio rows and rows without a provider id get none.
- `lib-dashboard-rail` reports the cards inside its viewport (`visibleCardsChanged`, an `IntersectionObserver` rooted at the track with ~one card of margin; every card counts as visible where the API is missing). Cards leaving the list are reported gone at once.
- `DashboardPortalLiveEpgPresenter` (component-provided) unions both rails' visible keys with the pinned hero and syncs the wanted set on change, on the 30 s tick, and on a display-offset change.
- `DashboardPortalLiveEpgService` (root) runs the queue: at most 2 requests in flight, 200 ms between starts (the numbers `EpgQueueService` already uses against real panels), one card per request, each answer published the moment it lands. Only wanted keys are dequeued, so a card scrolled past before its turn is never requested. Answers live 60 s, a failed portal rests 30 s, an ended programme is re-asked but not within 30 s, and a changed offset or XMLTV source set drops every answer.
- The card shows a shimmer placeholder only before its first answer; a refresh keeps the previous answer on screen. The page never waits: cards render immediately and fill in one by one.

M3U cards keep the single batched XMLTV lookup (one local SQLite query, already non-blocking).

## Tests

- `dashboard-portal-live-epg.util.spec.ts` (6): Xtream/Stalker entry building, radio/M3U/non-live rejection, map resolution, stop-time parsing.
- `dashboard-portal-live-epg.service.spec.ts` (8, fake timers): per-card incremental answers, ≤2 in flight with spacing, dropped-when-scrolled-away, TTL, ended-programme refetch with floor, failure cooldown, offset retirement, source-change re-ask.
- `dashboard-portal-live-epg.presenter.spec.ts` (4): pinned ∪ visible deduped across rails, stale keys ignored, offset re-sync, answer/pending reads.
- `dashboard-rail.component.spec.ts` (+4): observer rooted at the track with margin, item-order emission and no duplicate emits, re-observe on item change with removed cards reported gone, no-observer fallback, placeholder rendering.
- Electron E2E `dashboard-activation.e2e.ts`: the favourited Xtream channel's card in both live rails shows the mock's programme on air. Verified to fail with the presenter disconnected (`Received: " Xtream · Live "`).
- `pnpm nx test workspace-dashboard-feature` 72/72, `workspace-dashboard-data-access` (new suites 15/15), lint for both + `electron-backend-e2e`, `pnpm run typecheck:ci`, `release:notes:validate`.

Not covered by E2E: Stalker (the Electron suite has no Stalker favourites flow yet); its path is the same resolver call and is unit-covered by the entry builder.

## Docs

- `.changes/dashboard-portal-live-epg.md` (feature).
- `docs/architecture/workspace-dashboard.md` — Data Flow item 3 describes the lazy portal path; `CLAUDE.md` EPG section has a one-line pointer.


4gray:
All four findings fixed in 39ad7f0f5.

**Greptile P1 / Codex P2 — stale answers across a source change.** Valid, and the repo's late-result invalidation contract. A request now captures `EpgSourceSettingsService.revision()` next to the display offset and requeues itself when either moved. The requeue has to happen in the completion: while the key is in flight the retire pass cannot queue a replacement, so without it the answer computed against the removed guide was published and trusted for a full TTL.

**Greptile P2 / Codex P2 — wanted keys survive teardown.** Valid: the queue lives in the root service. The presenter now hands its wanted set back on destroy (`sync([])` via `DestroyRef`).

**Codex P2 — failures cached as empty EPG.** Valid, the cooldown was dead code: `loadEpgForItems` catches per-channel portal failures and files them as `null`, so the outer catch never fired. Since the resolver cannot distinguish a dead portal from a guide-less channel, the cooldown is gone and an answer with no programme now expires after 30 s instead of 60 s — that is what lets a transient outage recover on the next tick. A resolver-level throw takes the same path.

**Codex P2 — PWA.** Valid. `sync()` now returns immediately without the local EPG program-lookup capability, so the PWA no longer queues work that can never produce an answer. Lifting the resolver's own gate for portal lookups would change the collection pages too, so that stays out of this PR; the release note and the architecture doc now say desktop.

New coverage: source-revision discard-and-requeue, presenter teardown, the shorter empty-answer TTL, a rejected resolver call, and the capability no-op. `workspace-dashboard-data-access` 18/18, `workspace-dashboard-feature` 73/73, `dashboard-activation.e2e.ts` green, lint + `typecheck:ci` clean.

4gray:
@codex review

4gray:
@greptileai

4gray:
Rebased onto master now that #1637 has landed.

#1637 introduced `DashboardLiveEpgPresenter` for the rails' XMLTV lookup, so rather than wiring a second presenter into the page component, the portal lookup hangs off that one:

- `DashboardLiveEpgPresenter` is the single facade the rails talk to. It derives the portal rows from the enabled rails itself, pins the hero key, forwards each rail's visible set to `DashboardPortalLiveEpgPresenter`, and `enrich()` returns the cards with their "now on air" row filled in — a portal answer first, the XMLTV title match as fallback, and the shimmer placeholder only before a card's first portal answer.
- The page component only provides both presenters, sets `liveEpgSourceKey` on its live cards, and forwards visible cards.

That leaves `workspace-dashboard-rails.component.ts` at **835 lines, seven fewer than before this branch**, instead of the ~60 it would have gained on a file the max-lines baseline covers. `generate-max-lines-baseline.mjs` produces an unchanged baseline.

Everything else is unchanged from the reviewed version: the bounded queue, the source-revision guard, the teardown release, the empty-answer TTL and the desktop-only gate.

Verification after the rebase: `workspace-dashboard-feature` 82/82 (the XMLTV presenter spec now stubs the portal one and covers the merge: portal answer wins, `null` falls back to XMLTV, placeholder only before the first answer), `workspace-dashboard-data-access` green, lint for both, `typecheck:ci`, `dashboard-activation.e2e.ts` green, and `epg.e2e.ts` 13/13 to confirm #1637's own coverage still passes on top.

4gray:
@greptileai

4gray:
@codex review

4gray:
Follow-up in 1a2e25645: the consolidation derived the pinned key from the first entry of the portal source list, which is a favourite when the hero rail is hidden — pinning that would keep asking for a card nobody can see. The hero row is now its own computed, gated on the rail exactly like the source list, with a spec covering both states.

`workspace-dashboard-feature` 83/83, lint and `typecheck:ci` clean.

4gray:
@codex review

4gray:
@greptileai

4gray:
@greptileai

PR #1648, opened 2026-09-20
fix(tmdb): a new series no longer matches its older, better-known namesake

4gray:
## The bug

A recent local-language series from a portal catalog rendered the poster, cast, genres and rating of an unrelated, much older foreign show.

The app's own cache records the shape of the lookup it made:

```
tv | title:<localized name>|year:2026|v3 | ru-RU | → <older show's id>
```

Title and year were both correct. TMDB returns titles in the **request** language, and the search is issued with `ru-RU` for Cyrillic queries, so two unrelated results came back carrying the very same `name`:

| candidate | `name` (ru-RU) | `original_name` | premiere | votes |
| --- | --- | --- | --- | --- |
| the catalog's item | *same localized name* | *same, it is the original* | 2026-07-24 | 4 |
| an unrelated foreign show | *same localized name* | *its own, non-Cyrillic* | 2018-12-14 | 26 |

`pickConfidentMatch` admitted the 2018 row through the series **"premiered earlier"** tolerance (`mediaType === 'tv' && year < wantedYear`) — portals report the running season's year while TMDB reports the premiere — and then handed every admitted candidate to `pickMostPopular`, which ranks on `vote_count`/`popularity` alone. The 2026↔2026 year agreement counted for nothing; 26 votes beat 4.

## Not an edge case

Because TMDB localizes titles, a non-English catalog collapses foreign shows onto ordinary local nouns. Running the real `cleanTitleForSearch` + `pickConfidentMatch` against live TMDB over **400 Cyrillic series titles sampled from a real catalog**:

- **20** normalized keys had a same-titled *older* series
- **16 of those** were the more popular row — i.e. the newer series would have been given the older show's metadata

## The fix

Rank admitted candidates by **year evidence first** (`yearEvidenceTier`), and let popularity break ties only *inside* the strongest tier any candidate reached:

1. the provider's exact year
2. a year off by one
3. *(tv only)* premiered earlier

The tolerance stays — it does real work. Three of the eight lookups that install actually made depend on it (asked 2026 / premiere 2025; asked 2025 / premiere 2023; asked 2026 / premiere 2023). But it is a last resort, not an equal.

**Accepted residual**, documented in the code and the architecture doc: a long-running show whose stated season year happens to *be* another same-titled show's premiere year now resolves to the newer show. Only the older show's season air dates could separate the two, and a search response does not carry them — while that shape needs three coincidences at once, against one that needs none.

Search cache keys move `|v3` → `|v4` with a matching `migration:tmdb-search-lookup-v4-cache-cleanup:v1` startup cleanup: a positive row naming the wrong show otherwise stays fresh for 30 days.

## Validation

**Live TMDB, patched matcher, replayed against every real lookup one install had cached — 8/8 resolved to the right show**, including the reported mismatch (previously the older namesake) and the three that rely on the "premiered earlier" tolerance.

**Tests added** (`tmdb-matcher.spec.ts`): the reported regression, the tolerance still winning when nothing matches the year, and exact-year beating off-by-one for movies. Updated: the retired-generation specs (`connection-migrations.spec.ts`, the real-SQLite-in-Electron `tmdb-search-cache-cleanup.spec.ts`) and the resolver's cache-key expectations.

**Commands run**
- `pnpm nx test services` — 165 passed
- `pnpm nx test database` — 44 passed (incl. the Electron real-SQLite migration run: skipped-release, previous-release, pre-person, fresh and repeated startups)
- `pnpm nx run-many --target=lint --projects=services,database` — clean
- `pnpm run typecheck:ci` — clean
- `pnpm run release:notes:validate` — 86 notes valid

No E2E: the change is a pure ranking function plus a cache-key generation, with no route, IPC or persistence-shape impact beyond the additive cleanup already covered by the real-SQLite spec.

**Docs**: `docs/architecture/tmdb-metadata-enrichment.md` (ranking contract + the `|v4` generation) and the TMDB section of `CLAUDE.md`. Release note: `.changes/tmdb-new-series-year-match.md`.


4gray:
@codex review

4gray:
@greptileai

4gray:
@codex review

PR #1656, opened 2026-09-21
fix(embedded-mpv): unbreak the Windows runtime pin and the refresh that abandoned it

4gray:
## The breakage

`Build on windows x64` fails repository-wide as soon as a runner's cache goes cold:

```
Unable to download Windows embedded MPV runtime archive: 404 Not Found
```

`tools/embedded-mpv/windows-runtime-pin.json` pointed at an upstream release published 2026-08-21 with `retentionDays: 30`. Upstream deleted it — the asset **and** the tag are both 404 now. Nothing to download, so a re-run cannot help.

## Why the automation did not prevent it

`refresh-windows-embedded-mpv-runtime.yaml` exists precisely to rotate the pin well before that boundary. It had been red since **2026-09-07** (last success 2026-08-31), failing in its own `Validate updated pin` step with an `AssertionError`.

The cause is a loop that closes on itself:

- `createPinFixture()` wrote a copy of the **checked-in** pin into the temp dir;
- `age threshold rotates an available pin before upstream retention` then asserted, against a clock hard-coded to `2026-09-05T13:00:00Z`, that *that* pin was at least `WINDOWS_RUNTIME_REFRESH_AFTER_DAYS` (14) old.

So every **successful** rotation invalidated the assertion guarding it: the updater wrote a fresh pin, the validation step asserted the fresh pin was two weeks stale, the step went red, the bot PR was never opened — and the pin it had declined to land was left to expire. Reproduced locally: with the freshly rotated pin in place the old test file fails `not ok 7`, and it is the only failure.

## The fix

**Tests exercise the rule, not production data.** `createPinFixture({ ageDays })` now builds its pin from the existing synthetic release fixture at an age the test picks, against a named `REFRESH_CLOCK`:

| case | age | expected |
| --- | --- | --- |
| `available pin one day under the threshold is left alone` | 13 d | `current`, file untouched, one HEAD request |
| `age threshold rotates an available pin before upstream retention` | exactly 14 d | `age-threshold`, rotated to the newest release |
| `unavailable pin rotates to the newest downloadable release` | 1 d | `unavailable` — young on purpose, so only the 404 can explain it |

That converts a production-data assertion into real boundary coverage, and the isolation is now explicit in the unavailable case rather than incidental.

**New case** — `a pin rotated moments ago does not immediately ask to rotate again` — pins the invariant the job actually depends on: its own freshly written pin must read as `current` on the next run, whatever date that run falls on. That is the property the old test violated.

`CURRENT_PIN` stays with the schema, naming and licence-statement checks, which validate the checked-in file and hold for any pin.

**Pin rotated** to the newest upstream LGPL x86_64 release, via the repo's own `pnpm embedded-mpv:windows-runtime-pin:refresh` (reason reported as `unavailable`, correctly). The binary stays on its upstream host — IPTVnator does not mirror it — and the limited checksum/layout-only licence statement is preserved verbatim by the updater.

## Validation

- `node --test tools/embedded-mpv/windows-runtime-pin.test.mjs` — **9/9 pass**
- Data-independence proven both ways: the new tests pass with the **old** pin checked in *and* with the new one; the **old** test file fails against the new pin (`not ok 7`). The regression is genuinely covered.
- `pnpm embedded-mpv:windows-runtime-pin:check` — valid
- Archive verified end-to-end: downloads (27 MB) and its SHA-256 matches the pinned digest byte for byte
- `pnpm run release:notes:validate` — 87 notes valid

**Docs**: `tools/embedded-mpv/README.md` records the trap — nothing in the refresh tests may read the checked-in pin's `publishedAt` — beside the commands it already documents.

**No release note**: CI plumbing under `tools/`, outside the gate's `apps/` + `libs/` trigger, with no user-visible behaviour change. The second commit carries a `.changes` edit that is itself a note correction, not a new note.


4gray:
@codex review

PR #1658, opened 2026-09-21
docs(tmdb): replace catalog titles in matching examples with stand-ins

4gray:
## Why

The Cyrillic and Arabic examples documenting title folding were taken straight from a user's own portal catalog while the folding bugs were being diagnosed. From there they spread into code comments, test fixtures, assertions, cache-key strings and the architecture doc — and the TMDB ids pinned alongside them identify the exact shows. A public repository is not the place for someone's viewing inventory.

## What changed

Stand-ins that reproduce the **property under test** rather than the title:

| was | now | property preserved |
| --- | --- | --- |
| two spellings differing by "й" | `Лейка` / `Леика` | both fold to one key, stay two different wire queries |
| a title with "ё" | `Ёжик 2010` → `Ёжик` | "ё" survives the wire query, folds in the key |
| a two-word Cyrillic title + season marker | `Тестовый Сериал s02` | season stripping with "й" intact |
| a Cyrillic miss | `Пробный Выпуск` | negative-cache key shape |
| a Stalker season-marker title | `Пример (2 сезон)` | lazy-VOD season parsing |
| an Arabic title with initial hamza | `AR\| أمثلة تجريبية` | hamza decomposes into a separate word: key `ا مثلة تجريبية` |
| real TMDB ids | synthetic `1011xx` | — |
| a channel fixture that read as a film title | a plain channel name | ё case-folding search |

Every replacement was run through the real `normalizeTitle` / `cleanTitleForSearch` **before** being written, so folded keys, wire queries and cache lookup keys keep exactly the shape the assertions describe.

## Deliberately left alone

The leading-tag rule's `"Akira | 1988"` / `"Coco | 2017"` / `"Момо | Momo"` examples in `title-normalization.util.ts`. Those are not an inventory — they are the evidence for a regex decision measured over 1.27M catalog titles ("corrupted 349 keys and rescued none"). Inventing replacements there would document a measurement that never happened.

## Validation

- `pnpm nx run-many --target=test --projects=services,database,shared-interfaces,portal-stalker-data-access,electron-backend --skip-nx-cache` — **all 5 pass** (includes the real-SQLite-in-Electron migration spec and the SQLite content-search spec, both of which carry affected fixtures)
- Prettier clean on every changed file
- Repository-wide grep confirms no remaining occurrence of the replaced titles or ids

`no-release-note`: comments, fixtures and documentation only. The two non-spec files under `libs/` (`title-normalization.util.ts`, `tmdb-matcher.ts`) and `connection.ts` change comment text only, so the gate's path trigger fires on a change with no user-visible behaviour.


4gray:
@codex review

4gray:
Both bots landed on the same finding, and it is right: the substitution left synthetic titles inside sentences asserting observed fact — the doc claimed a specific folded query "returns zero results while `Лейка` returns the show", and named the stand-ins as the titles that were searched, missed and negatively cached. Those titles never existed, so a future reader debugging a regression would take them for production evidence.

Fixed in f0096c643 across the doc and every comment that read the same way:

- the class-level historical claim is kept, because that is what was actually established — under the old single-form design every Cyrillic title carrying "й"/"ё" and every Arabic title carrying a hamza form was searched folded, missed, and cached as missing for the negative TTL;
- the strings are now explicitly illustrative stand-ins chosen to fold the same way, stated once in the doc section and again in the resolver spec header;
- claims about what TMDB indexes for a particular made-up string are gone (`tmdb-matcher.ts`, `connection.ts`, `title-normalization.util.ts`, and the two spec comments).

The verified invariant is untouched and still asserted: what NFD does to those letters and what the two tiers therefore produce is a property of the text, not of any title.

@greptileai

4gray:
@codex review

4gray:
Both findings were right; fixed in 278079519.

**Missed cache key** — `tmdb-search-cache-cleanup.spec.ts` kept a catalog-derived lookup key verbatim, twice. The cause is worth naming: the sweep matched titles in their display casing, and a lookup key stores the title lowercased, so only the TMDB id beside it got swapped. Replaced with a synthetic key, and the sweep was re-run case-insensitively over every original title and id — the repository is now clean.

**Observed-fact claim in `SearchTitleVariant`** — the comment still asserted that one specific folded string finds nothing on TMDB, which no test can support. It now states the mechanism: folding rewrites the letters the search matches on, so the folded form finds nothing there, while it is exactly what the confidence gate and the cache need.

One place is deliberately **not** changed, and it is worth recording why: `search-text-fold.util.spec.ts`. Its "Ёлки" is a common noun in a Unicode corpus beside "Amélie", "İnşaat" and "Ά" — not an inventory entry. More importantly its rows are precomposed/decomposed PAIRS (`U+0401` against `U+0415 U+0308`), and a textual substitution rewrites only the composed half, silently turning the pair into two different words. I tried it, the spec failed, and I reverted — which is exactly what a fixture encoding an invisible property is for.

Also left alone, for the same reason as before: the leading-tag examples in `title-normalization.util.ts`. Those are the measured evidence for a regex decision over 1.27M catalog titles, not an inventory; inventing replacements there would document a measurement that never happened.

Verification: `services`, `shared-interfaces`, `database`, `electron-backend` all pass.

@greptileai

4gray:
@codex review

4gray:
Done in 4313bb5ee — and you were right to push back, though the naive form of the change is a trap worth recording.

Those two rows are not plain strings. One of them is a **precomposed/decomposed pair**:

```
['Ёлки', 'Ёлки']
  U+0401 л к и      U+0415 U+0308 л к и
```

The test asserts that both spellings fold to one string. A textual substitution rewrites only the composed half, silently turning the pair into two different words — the two look identical in a diff, in a review and to `rg`. That is why the earlier attempt was reverted: it failed the spec (1 of 534).

Rewritten by code point instead, keeping the decomposed half decomposed (`U+0415 U+0308` + the new stem), verified by decoding every quoted string in the file afterwards. `shared-interfaces`: 534/534 pass.

A repository-wide sweep now finds no occurrence of the replaced titles in **either** normalization form — the plain grep you ran would not have caught a decomposed leftover, so that check is now part of the sweep.

Unrelated: `Build on linux x64 (portable)` failed on the previous revision in `embedded-mpv-frame-copy-packaged.e2e.ts`. The same job passed on both earlier revisions of this PR and the commit touched only a comment and two spec strings, so it was re-run and is green.

@greptileai

4gray:
@codex review

4gray:
@greptileai

